PR #124, opened 2014-05-06
Allow process checks to be configured

### LINKED ISSUE #112: Process metrics can't be configured
While there is a template for the [Process stats monitoring](https://github.com/DataDog/chef-datadog/blob/66c200b4cba8ae43bc620da11f3bdc6ed899539b/templates/default/process.yaml.erb) it doesn't have any parameterisation to let you specify what you want to monitor. It looks like it was added with no recipe/attributes to customise it.

Sinjo:
Adds templating to `process.yaml.erb` and a corresponding recipe to allow per-process stats to be configured. Fixes #112.

Also fixes a small issue in the documentation, as exact match [doesn't default to False](https://github.com/DataDog/dd-agent/blob/4cb1a7226bfcfcaa9fb175e4fcbbb4b4eabbc7b7/checks.d/process.py#L30)

### DIFF
--- a/recipes/process.rb
+++ b/recipes/process.rb
@@ -1,0 +1,29 @@
+include_recipe "datadog::dd-agent"
+
+# Monitor processes
+#
+#    name:          (mandatory) STRING Is the display name of the check
+#    search_string: (mandatory) LIST OF STRINGS The strings to search for in process names. If
+#                               one of these matches a process's name, that process will be
+#                               included in the stats.
+#    exact_match:   (optional)  True/False, defaults to True, if you want to look for a partial
+#                               match, use exact_match: "False", otherwise use the exact base
+#                               name of the process.
+#
+# Example:
+#
+# node.datadog.process.instances = [
+#   {
+#     :name => "ssh",
+#     :search_string => ["ssh","sshd"],
+#     :exact_match => "False",
+#   },
+#   {
+#     :name => "postgres",
+#     :search_string => ["postgres"],
+#   },
+# ]
+
+datadog_monitor "process" do
+  instances node['datadog']['process']['instances']
+end
--- a/recipes/process.rb
+++ b/recipes/process.rb
@@ -1,4 +1,4 @@
-include_recipe "datadog::dd-agent"
+include_recipe 'datadog::dd-agent'
 
 # Monitor processes
 #
@@ -24,9 +24,9 @@ include_recipe "datadog::dd-agent"
 #   },
 # ]
 
-package "python-setuptools"
-easy_install_package "psutil"
+package 'python-setuptools'
+easy_install_package 'psutil'
 
-datadog_monitor "process" do
+datadog_monitor 'process' do
   instances node['datadog']['process']['instances']
 end
--- a/recipes/process.rb
+++ b/recipes/process.rb
@@ -24,6 +24,9 @@ include_recipe "datadog::dd-agent"
 #   },
 # ]
 
+package "python-setuptools"
+easy_install_package "psutil"
+
 datadog_monitor "process" do
   instances node['datadog']['process']['instances']
 end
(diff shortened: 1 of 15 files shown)


Sinjo:
@miketheman Happy to figure out how this fits into 5.x and update the recipe. Before I do that though, I'd like to look deeper at some issues we ran into.

We found the process monitoring was of limited use because of the permissions set on certain stats in procfs. For example, the number of bytes read/written by a process could only be read by that process, so the only process which got really detailed stats was the stats collecting process itself.

I've been struggling to come up with a solution. I'm not sure if it's possible to configure those permissions, or whether that would even be sensible. Running the stat collector as root doesn't seem like a reasonable approach either.

Any thoughts?

miketheman:
Hi @Sinjo,
I wonder if you're running into an issue that can be solved with the correct flag for that process:
https://github.com/DataDog/dd-agent/blob/367819883e3bea937fd178a14a0710ececc3ce7e/conf.d/process.yaml.example#L10-L11

As discussed here: https://github.com/DataDog/dd-agent/issues/866 The only other option for some of those metrics is to run the Agent as root, or add the Agent to the root group to allow access - not something we particularly recommend.

So I think proceeding with the process.yaml.erb update makes sense - so let me know if you'd like to continue it, or if we should pick it up.

Sinjo:
Wasn't aware of the flag to suppress the logging, but that wasn't our main issue. The stats we were hoping to trace when I wrote this ended up being the ones which required root (and I agree with you, I wouldn't want to run the Agent as root).

Happy to let someone else pick up this work if there's still value in it with the reduced set of stats (would be really good to document which ones work without root as well).

joelnn:
I just came across this, and would like to suggest -- it would be ideal if process checks could be configured via LWRP rather than a recipe.

Whether or not a process should be monitored is dependent on whether its installed, and it would be best if e.g. the postgres cookbook could add the postgres process check.

miketheman:
@joelnn I agree that would be ideal. When we started this, we deliberated on how that might work, since currently you can have multiple cookbooks wanting to monitor different processes, and there's only a single process.yaml file - how do we safely govern adding/removing entries into that file without clobbering each other and restarting the Agent?

So we left it up to the end-user to determine how they wanted to build the attributes and then call the `datadog_monitor` LWRP via the dd-agent:process recipe - so technically it still is an LWRP, just wrapped in a recipe for convenience.

jpcallanta:
@miketheman Hi, I'm wondering if there's an update on the status of this thread regarding process.yaml.erb. I've forked and started looking into implementing a process.rb recipe as it is needed in my organization. Is there still interest and value into getting this recipe merged in?

miketheman:
@jpcallanta there is! I've been working on the process and http/tcp checks for the next revision (along with system.core monitoring). I hope to have this out soon.

jpcallanta:
@miketheman curious if the implementation you are working on for the process recipe is along the lines of https://github.com/jpcallanta/chef-datadog/blob/master/recipes/process.rb and https://github.com/jpcallanta/chef-datadog/blob/master/templates/default/process.yaml.erb which I believe was the proposed changes on this thread. If so we can roll with these until the next version.

miketheman:
@jpcallanta that is a preferred method - might you consider adding a text-kitchen test for that and creating a PR? I think that is likely to ease the process (pun!).

jpcallanta:
@miketheman sure thing I'll work on writing the test.

miketheman:
@jpcallanta @mtougeron - I think y'all might be battling over the process check - it'd be helpful to have the "simpler" rendering interface with tests.
See https://github.com/DataDog/chef-datadog/pull/181 as well.

mtougeron:
Either way works for me. I just need it. :)

recipes/process.rb
```diff
@@ -0,0 +1,32 @@
...
+#     :exact_match => "False",
+#   },
+#   {
+#     :name => "postgres",
+#     :search_string => ["postgres"],
+#   },
+# ]
+
+package 'python-setuptools'
+easy_install_package 'psutil'
```
mtougeron: I defer to @miketheman but isn't Python supposed to be self-contained for Datadog?

mtougeron:
@miketheman actually as I look a little closer at this, is there any reason why my changes couldn't be merged while @jpcallanta finishes up his? PR #181 adds most of the missing code that you are looking for in this one. Then this PR would just be about the addition of psutil. Thoughts?

miketheman:
Thanks gang! Merged in f048bbb62b16f69cfa1fca06228b3a93fc7152f4 which should address process checks.

PR #138, opened 2014-08-22
Update keyserver url for debian OS

khouse:
We were seeing timeouts connecting to this url for our VPC instance builds, and the issue was resolved by doing explicit port settings with the hkp protocol. Hopefully you will consider this change!

http://superuser.com/questions/64922/how-to-work-around-blocked-outbound-hkp-port-for-apt-keys


miketheman:
Thanks!

Issue #96, opened 2014-01-30
Support for encrypted data bags


evan2645:
The readme indicates that data bags were purposefully left out under the assumption that it's unlikely someone would have more than one api key. What has been overlooked is that many people prefer to store credentials of any kind (including api keys) in an encrypted data bag to avoid keeping sensitive information in node attributes.

I'm more than willing to open a PR with the necessary modifications to optionally support an encrypted data bag, but would like to gauge interest first. Would you accept such a feature?

PR #149, opened 2014-10-22
add encrypted databag support

shortdudey123:
Updated old pull request to the newest head revision (https://github.com/DataDog/chef-datadog/pull/99)

Solves https://github.com/DataDog/chef-datadog/issues/96

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -44,6 +44,18 @@ suites:
       api_key: somethingnotnil
       application_key: alsonotnil
 
+- name: dd-agent-encrypted-data-bag
+  data_bags_path: test/integration/dd-agent/data_bags
+  encrypted_data_bag_secret_key_path: test/integration/dd-agent/encrypted_data_bag_secret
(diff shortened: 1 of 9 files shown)


shortdudey123:
@miketheman that assumes the infrastructure is already running Chef 12.

rm does not exist in Cheff 11 and prior
http://www.rubydoc.info/gems/chef/11.16.4/Chef/Node/Attribute

ranjib:
@miketheman may be this is not the right approach, but setting a secret as raw node attribute, which is searchable by all other nodes (or any non-admin chef api client) is really a security hole. Not only we should address that (i.e. offer any other mechanism that does not leak the password), we should also declare all the resources (like template) that uses the creds as sensitive.

miketheman:
Reviewed again. We won't fix this right now, and will explore other security measures for this in the next major release.

shortdudey123:
no problem, i will keep the branch around in case you want to use it to improve security in the future :)

PR #152, opened 2014-10-29
adding mesos

JohnFiedler:
Adding mesos yaml and recipe per datadog instruction https://app.datadoghq.com/account/settings#integrations/mesos


miketheman:
Hey @JohnFiedler @ Thanks for taking the time, we merged in #200 which implements this instead.

Issue #158, opened 2014-12-04
cannot set ec2_collect_tags via chef aws opsworks install


mirceal:
I have 2 opsworks stack and in order for me to differentiate between them I need to be able to collect the ec2 tags associated with the machine.
if I manually set 
collect_ec2_tags: no
in /etc/dd-agent/datadog.conf I can see the tags being collected.

When I run the chef recipe via opsworks though, I don't see this parameter being configured in datadog.conf 
Looking at the template erg seems that the param is not there.

Added ['datadog']['collect_ec2_tags'] and ['datadog']['api_key'] to the custom json used by opsworks and the api_key is picked up just fine (i see it ending up in config) but the collect_ec2_tags is just ignored.

Need to find a way to add collecting the ec2 tags in config when the agent is installed via chef (or an workaround)

miketheman:
PR has been merged, missed the link to this issue. Closing!

xbeta:
@miketheman when can we expect newer release with this change?  We are at v2.0.0 now

miketheman:
Targeting April 20th.

PR #159, opened 2014-12-04
add option to specify ec2 tag collection

mirceal:
issue opened on chef-datadog: https://github.com/DataDog/chef-datadog/issues/158

I have 2 opsworks stack and in order for me to differentiate between them I need to be able to collect the ec2 tags associated with the machine.
if I manually set 
collect_ec2_tags: no
in /etc/dd-agent/datadog.conf I can see the tags being collected.

When I run the chef recipe via opsworks though, I don't see this parameter being configured in datadog.conf 
Looking at the template erg seems that the param is not there.

Added ['datadog']['collect_ec2_tags'] and ['datadog']['api_key'] to the custom json used by opsworks and the api_key is picked up just fine (i see it ending up in config) but the collect_ec2_tags is just ignored.

Need to find a way to add collecting the ec2 tags in config when the agent is installed via chef (or an workaround)


miketheman:
@mirceal I'm re-reading this, and I think the condition that you are applying will _always_ default to being a truthy value, and render the `collect_ec2_tags` template line, even if the default is `'no'`, since the value of the attribute evaluates to a valid String (and not nil).

So it may make sense to provide a default of `nil`, and comment that this should be a valid String of `yes` to activate the collection.

mirceal:
yeah.
there was a bit of confusion on my side when I needed to figure a default value for this.
I did look at the agent before deciding on 'no':
https://github.com/DataDog/dd-agent/blob/c4118836ede5a044c5ede8a029d63d2986996352/config.py#L313

mirceal:
also, as a side note: this isn't as important as I thought it to be went first started with data dog.
because I was a noob and did not really get the way the data dog integrations work, I did not turn the AWS integ off the bat (and tried pulling instance tags using this method). If the AWS integ is turned on, it will do the right thing as far ar pulling the tags and matching the hosts with AWS instances.

miketheman:
@mirceal That makes total sense - thanks for the link to the source.
I think the `if/end` condition is now not really useful, since the flag will always be rendered, just with the default of `no`, which in turn, will be False in the Agent.
So that could be simplified a bit.

Knowing that the use of the AWS Integration solves the problem as a whole, should this PR still be considered? Or does adding it provide potential for further confusion, since it's under the DEPRECATED section anyhow since 2012. https://github.com/DataDog/dd-agent/commit/3073c60c80b3a2d9f4557d1f6e12d1c3b32e8ba3

mirceal:
this enables using the agent on AWS without enabling the data dog integration (which may be of some value).

mirceal:
yeah... figured that out. take a look now :)

mirceal:
ping

miketheman:
@mirceal Thanks - looking at the changes, I still don't see the removal if the `if/end` condition - the existence of anything in https://github.com/DataDog/chef-datadog/pull/159/files#diff-25e5d4a4446ae12a0d6f1162b6160375R37 no/yes/foobar negates the condition, so it's not needed.

mirceal:
guess I misunderstood what you were asking.
the condition was there to make it uniform with the way the config was written (i.e. consistency not need made me put it there)

miketheman:
In that case, the default option in the attributes should be `nil` - so as to not write anything if the condition is not set.
If someone sets it to explicitly `no`, then it'll get written.
Does that make sense?

mirceal:
yes it does make sense :)

mirceal:
updated

miketheman:
Thanks! I'll likely merge soon.

mirceal:
thanks & thanks for the patience

miketheman:
Merged (with a bit of rebasing and docs updates). Thanks!

PR #163, opened 2014-12-18
Update kafka.yaml: proper spacing & recent changes

donaldguy:
With previous spacing check throws exception:
  InvalidJMXConfiguration("Each configuration must have an 'include' section.")

In particular incorporates
DataDog/dd-agent@75863d82ee2f0cdef3f3e75097fe79cb17f6b5c2 - counters-> guages,
DataDog/dd-agent@5382422d3790be2a9c4f6eccb9bca58547ae8c76 - jmxfetch custom tags
and DataDog/dd-agent@8bac88e23f94d615a90e587ed84c55e4adba683f - support for jmxfetch0.4

### DIFF
--- a/templates/default/kafka.yaml.erb
+++ b/templates/default/kafka.yaml.erb
@@ -11,6 +11,11 @@ instances:
     <% if i['password'] -%>
     password: <%= i['password'] %>
     <% end -%>
+    <% if i['process_name_regex'] and i['tools_jar_path'] -%>
+    process_name_regex: <%= i['process_name_regex'] %> # Instead of specifying a host, and port. The agent can connect using the attach api.
+                                                       # This requires the JDK to be installed and the path to tools.jar to be set below.
+    tools_jar_path: <%= i['tools_jar_path'] %>
+    <% end -%>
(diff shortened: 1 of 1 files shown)


donaldguy:
I'm a little confused about appropriate scope

What in particular do you think should be tested? Should I constrain it to testing the contents of the yaml file or should it encompass setting up a node with a toy kafka broker as well?

Put another way, there are existing bats tests for the kafka; why are they insufficient?

--
Seperately there are probably other JMX metrics worth gathering - in particular I probably want to alert on under-replicated partitions.  (and maybe encompass all that the LinkedIn folks highlight in https://kafka.apache.org/documentation.html#monitoring ); could I add these here, or do they need to get back to the dd-agent repo ? or should I just aim at an `other_beans` attribute or something?

CC: @remh @conorbranagan on the latter

miketheman:
HI @donaldguy!
Thanks for the contribution - as you can see, your patch was merged in to master. I've added the serverspec tests style that will pass in the test-kitchen minimal environment - that's what I was referring to - bats doesn't always cut it and is harder to maintain.

Issue #165, opened 2014-12-19
dd-handler can't send events on Windows due to missing SSL_CERT_FILE env var


rlaveycal:
The openSSL gem used by the dd-handler requires that the SSL_CERT_FILE environment variable is set so that it can find the certificate store to validate the server's cert. On Windows this is not set by default so the handler fails.

In Chef 11.16, SSL_CERT_FILE is set for Chef components (see https://github.com/opscode/omnibus-chef/pull/236) but the dd-handler doesn't see/use it.

err0:
https://github.com/chef/chef-dk/issues/106 had this issue too.

bigbam505:
I added documentation on the dd-hadler site you can fix this with the following:

https://github.com/DataDog/chef-handler-datadog#windows-support

miketheman:
Closing, in favor of documentation update.

PR #166, opened 2014-12-19
Set SSL_CERT_FILE in Windows

rlaveycal:
This is a Quick fix for #165. Better would be to use the value provided through Chef


miketheman:
Closing as this was resolved via better documentation (note https://github.com/DataDog/chef-datadog/issues/165#issuecomment-150365986)

See https://github.com/DataDog/chef-handler-datadog#windows-support

PR #167, opened 2014-12-20
Add support for Postfix integration

phlipper:
This PR introduces support for the Postfix integration.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -349,3 +349,25 @@ suites:
           - name: test2
             host: localhost
             port: 5678
+- name: datadog_postfix
+  run_list:
+    - recipe[sudo]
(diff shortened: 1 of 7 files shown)


phlipper:
This has been rebased on top of the latest master, and contains only the postfix changes.

miketheman:
@phlipper Thanks! I'd like to take one pass at seeing if we can get the template syntax down to the simpler, new template model: https://github.com/DataDog/chef-datadog/blob/master/templates/default/ntp.yaml.erb

This would allow for arbitrary extensions to the plugin to be used immediately by manipulating the attribute hash, instead of waiting for a cookbook release to the template.

miketheman:
Merged, thanks for working on this!

PR #168, opened 2014-12-20
Fix Provider `updated_by_last_action` return value

phlipper:
This PR uses the return value from the executed resource vs. hard-coding a `true`/`false` return.

### DIFF
--- a/providers/monitor.rb
+++ b/providers/monitor.rb
@@ -6,25 +6,29 @@ end
 
 action :add do
   Chef::Log.debug "Adding monitoring for #{new_resource.name}"
-  template "/etc/dd-agent/conf.d/#{new_resource.name}.yaml" do
+
+  t = template "/etc/dd-agent/conf.d/#{new_resource.name}.yaml" do
     owner 'dd-agent'
-    mode 00600
+    mode '0600'
     variables(
-      :init_config => new_resource.init_config,
-      :instances   => new_resource.instances
+      init_config: new_resource.init_config,
+      instances: new_resource.instances
     )
-    notifies :restart, 'service[datadog-agent]', :delayed
+    notifies :restart, 'service[datadog-agent]'
   end
-  new_resource.updated_by_last_action(false)
+
+  new_resource.updated_by_last_action(t.updated_by_last_action?)
 end
 
 action :remove do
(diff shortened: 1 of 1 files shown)


miketheman:
This commit does a lot more than update the action, it changes style as well.
Not saying that the change isn't welcome, but sectioning them to behavior vs style changes is a better approach, as they can be diff'ed accordingly.

phlipper:
@miketheman are you asking me to change this PR now, or are you just commenting generally?

miketheman:
I think I'm commenting in general. i've been working through a bunch of bug tickets here, and commenting on issues as I notice things.
I may split & rebase the code so that it does that when I get to this feature.

olivielpeau:
@phlipper Unfortunately this approach doesn't change the current behavior because `t.updated_by_last_action?` is evaluated during the compile phase and therefore always evaluates to `false`. For reference see https://github.com/chef/chef/issues/3748.

A comment on the above issue recommends using `use_inline_resources` instead, but it has the drawbacks of only working on Chef >= 11 and of not allowing external resources to be notified inside of the LWRP (see https://docs.chef.io/lwrp.html#disable). As the behavior of `use_inline_resources` will be the default one in future versions of Chef we should think about handling it though (maybe by notifying the `datadog-agent` service when `datadog_monitor` is used, or by using [definitions](http://docs.chef.io/definitions.html#many-recipes-one-definition)).

Also, for reference, this [email](http://lists.opscode.com/sympa/arc/chef/2015-06/msg00435.html) on the chef mailing list recommends running the embedded resource of the LWRP before evaluating `updated_by_last_action?` on it, but this approach has the same drawback of not allowing to notify external resources in the embedded resource.

@miketheman Any thoughts on this?

miketheman:
@olivielpeau The current cookbook aims to support Chef 10, so a pattern used there is:
```ruby
use_inline_resources if defined?(use_inline_resources)
```
This pattern [dates back a while](https://tickets.opscode.com/browse/COOK-3225).

So I think the final desire could be accomplished using the inline resources approach, as long as it's qualified.

There's also [this post](http://jtimberman.housepub.org/blog/2015/01/17/quick-tip-define-resources-to-notifiy-in-lwrps/) on the approach.

PR #169, opened 2015-01-02
Add support for a few more Elasticsearch integration options

chrissnell:
Adds support for is_external, username, password, and tags.


miketheman:
Looks good! There's a one-space off in the `tags` section around the `if/end` statements - can you fix that?

chrissnell:
Hi @miketheman , does this new commit fix what you're talking about?

miketheman:
@chrissnell no, I was referring to the 3-spaces on the if/end - it should be 2 - lines 15 & 17

chrissnell:
Ahh, I see now.  Better?

chrissnell:
OK, I added a space to make it a three-space to match the rest of the spacing in the template.  If you want two-space everywhere, LMK and I will change everything to match.

miketheman:
I think mentioned that it should be 2 in my prior comment: https://github.com/DataDog/chef-datadog/pull/169#issuecomment-68553085

This is how I would expect to see the tags block indented:
```erb
    tags:
      <% i['tags'].each do |t| -%>
      - <%= t %>
      <% end -%>
    <% end -%>
```

3-spacing is a very odd one - that why I pointed it out. 2-spacing is pretty standard Ruby and YAML.

Would you mind rebasing and squashing these commits down to a single one?

chrissnell:
\o/

miketheman:
Nice! I'll merge this soon with the next round of enhancements.

PR #171, opened 2015-01-05
Fix YAML formatting

SelerityMichael:
YAML had incorrect formatting in the Ruby template, causing the following error after Chef used it to create the `activemq.yaml` config file:

```
    activemq
    --------
      - initialize check class [ERROR]: InvalidJMXConfiguration("Each configuration must have an 'include' section. See http://docs.datadoghq.com/integrations/java/ for more information",)
```


miketheman:
Thanks!

Issue #141, opened 2014-09-19
default kafka.yaml doesnt seem to work


jchauncey:
Here's the yaml file created by the cookbook. It seems to be emitting system metrics but not jmx metrics.. Any suggestions"

```yaml
instances:
  - host: localhost
    port: 9920

init_config:
  is_jmx: true

  # Metrics collected by this check. You should not have to modify this.
  conf:
    #
    # Aggregate cluster stats
    #
    - include:
      domain: '"kafka.server"'
      bean: '"kafka.server":type="BrokerTopicMetrics",name="AllTopicsBytesOutPerSec"'
      attribute:
        MeanRate:
          metric_type: counter
          alias: kafka.net.bytes_out
    - include:
      domain: '"kafka.server"'
      bean: '"kafka.server":type="BrokerTopicMetrics",name="AllTopicsBytesInPerSec"'
      attribute:
        MeanRate:
          metric_type: counter
          alias: kafka.net.bytes_in
    - include:
      domain: '"kafka.server"'
      bean: '"kafka.server":type="BrokerTopicMetrics",name="AllTopicsMessagesInPerSec"'
      attribute:
        MeanRate:
          metric_type: gauge
          alias: kafka.messages_in

    #
    # Request timings
    #
    - include:
      domain: '"kafka.server"'
      bean: '"kafka.server":type="BrokerTopicMetrics",name="AllTopicsFailedFetchRequestsPerSec"'
      attribute:
        MeanRate:
          metric_type: gauge
          alias: kafka.request.fetch.failed
    - include:
      domain: '"kafka.server"'
      bean: '"kafka.server":type="BrokerTopicMetrics",name="AllTopicsFailedProduceRequestsPerSec"'
      attribute:
        MeanRate:
          metric_type: gauge
          alias: kafka.request.produce.failed
    - include:
      domain: '"kafka.network"'
      bean: '"kafka.network":type="RequestMetrics",name="Produce-TotalTimeMs"'
      attribute:
        Mean:
          metric_type: counter
          alias: kafka.request.produce.time.avg
        99thPercentile:
          metric_type: counter
          alias: kafka.request.produce.time.99percentile
    - include:
      domain: '"kafka.network"'
      bean: '"kafka.network":type="RequestMetrics",name="Fetch-TotalTimeMs"'
      attribute:
        Mean:
          metric_type: counter
          alias: kafka.request.fetch.time.avg
        99thPercentile:
          metric_type: counter
[...]

PR #176, opened 2015-01-26
Fix to kafka.yaml.erb indentation

solarce:
- 
  Without this kafka.rb is generate a kafka.yaml that is not valid

For #141.


donaldguy:
See also #163 - there were some other changes in the agent repo that I ported there

PR #182, opened 2015-02-12
New NTP recipe

chrissnell:
A recipe for the NTP check.   Uses the new style Ruby->JSON->YAML configuration translation.

### DIFF
--- a/recipes/ntp.rb
+++ b/recipes/ntp.rb
@@ -1,0 +1,17 @@
+include_recipe 'datadog::dd-agent'
+
+# Build a data structure with configuration.
+# @see https://github.com/DataDog/dd-agent/blob/master/conf.d/ntp.yaml.example
+# @example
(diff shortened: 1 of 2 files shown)


miketheman:
Nice work! Any chance of some tests?

miketheman:
Hey @chrissnell ,

Getting back around to this one, I think it looks good, however the Agent team has actually determined that this is one of those checks that should be enabled by default. See https://github.com/DataDog/dd-agent/pull/1433 for more.

This should be making it into the 5.3.0 release in a few weeks, so I'm reluctant to add this to the mix right now - rather once 5.3.0 comes out, it's on by default for anyone.

How does that sound to you?

miketheman:
Hi @chrissnell !
Sorry for any confusion, got this rebased and merged, along with a test. Thanks!

Issue #170, opened 2015-01-02
Shouldn't node['datadog']['tags'] be an array or hash?


chrissnell:
In the datadog.conf.erb template, I see that node['datadog']['tags'] is a string.  How do y'all feel about changing this attribute to an array or hash so that tags could be added by other cookbooks dynamically without overwriting tags set elsewhere.   The template could just do a .each to build the tag line in the configuration file.

Presumably the template could check the type of the attribute and do the right thing if the attribute is a string to retain backwards compatibility.

miketheman:
Hey @chrissnell,

You bring up a good point.

I guess we didn't look at the use case too hard back when it was submitted.
One _could_ go about appending to a String, but that might not be great with commas and such.

I'm curious to see what you would have in mind.

awasilyev:
Any updates?

miketheman:
https://github.com/DataDog/chef-datadog/pull/186 Has been submitted, there's some progress there - but it's not ready for merge/release yet.

miketheman:
Resolved via #296.

chrissnell:
Did this ever get released?

olivielpeau:
@chrissnell This got released with `v2.3.0` (see https://github.com/DataDog/chef-datadog/blob/v2.3.0/attributes/default.rb#L39-L45)

PR #183, opened 2015-02-17
Support for multiple tags in datadog.conf

lightpriest:
Enables setting tags from different attribute levels, backward compatible.

Suppose you have role A which acts on two environments B and C.
This patch enables setting this on the role:

    {
      "datadog": {
        "tags": {
          "cluster": "A"
         }
       }
    }

And these on the different environments:

    {
      "datadog": {
        "tags": {
          "environment": "B"
         }
       }
    }

    {
      "datadog": {
        "tags": {
          "environment": "C"
         }
       }
    }

Will result in datadog.conf having `tags: cluster:A,environment:B` and `tags: cluster:A,environment:C` respectively. String version of the tags should still work.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -30,7 +30,7 @@ default['datadog']['application_key'] = nil
 # The host of the Datadog intake server to send agent data to
 default['datadog']['url'] = 'https://app.datadoghq.com'
 
-# Add tags as override attributes in your role
+# Add tags as override attributes in your role, string or key/value
 default['datadog']['tags'] = ''
 
 # Collect EC2 tags, set to 'yes' to collect
--- a/metadata.rb
+++ b/metadata.rb
@@ -4,7 +4,7 @@ maintainer_email 'package@datadoghq.com'
 license          'Apache 2.0'
 description      'Installs/Configures datadog components'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version          '2.0.0'
+version          '2.1.0'
 
 %w(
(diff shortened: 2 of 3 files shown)


miketheman:
@lightpriest Thanks for working on this. I think this was originally reported in #170 

I think that the test for `respond_to?` may be misleading, as an Array also has a `map` method - and when iterating on it, you may have unexpected results.
```ruby
[1] pry(main)> ['a'].respond_to? 'map'
=> true
[2] pry(main)> ['a'].select{|k,v| not v.nil?}.map{|k,v| "#{k}:#{v}"}.join(",")
=> ""
```

So there may be some other item you may want to test - maybe something like `attribute.class == Hash` or such, or handling if it's an Array, etc.

Also, since we don't release a cookbook change for each feature, please don't increase the version number when sending a pull request - it's fine to maintain that on your own fork, but that's a definite blocker for merging.

lightpriest:
You're right, testing for `map` is misleading, but testing for the specific class could be even more problematic. Chef's attribute direct classes are `Chef::Node::ImmutableMash` and `Chef::Node::ImmutableArray`.

Though it should work, it might break in the future due to refactoring or some other implementation change in Chef. Testing for `respond_to?` is basically testing for the capability we're looking for, and `map` is confusing so I guess it's not quite what I'm looking for, maybe `each_pair`? Maybe I'm way off and there's a better way.. I'm open to suggestions.

I'm sorry for the version commit, I will create a new branch without it and will request a new pull.

PR #185, opened 2015-02-27
adding more redis config options supported by redis check

opsline-radek:
Adding more redis configuration options that are now supported by redis check. 

* db - to specify redis database for an instance
* keys - to get data for specific keys
* warn_on_missing_keys - to enable warning on the info page when defined keys are missing
* unix_socket_path - to connect using unix socket (disabled host and port)


miketheman:
Merged, thanks!

PR #186, opened 2015-03-02
Add support for multiple tags via key/value

lightpriest:
A new pull request (previous is #183) without a version change and with a more strict test for Hash implementation.

---

Enables setting tags from different attribute levels, backward compatible.

Suppose you have role A which acts on two environments B and C.
This patch enables setting this on the role:

    {
      "datadog": {
        "tags": {
          "cluster": "A"
        }
      }
    }

And these on the different environments:

    {
      "datadog": {
        "tags": {
          "environment": "B"
        }
      }
    }

    {
      "datadog": {
        "tags": {
          "environment": "C"
        }
      }
    }

Will result in datadog.conf having `tags: cluster:A,environment:B` and `tags: cluster:A,environment:C` respectively. String version of the tags should still work. To remove a tag in a higher precedence level, set it to an empty string.

This is also reported in #170.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -31,6 +31,8 @@ default['datadog']['application_key'] = nil
 default['datadog']['url'] = 'https://app.datadoghq.com'
 
 # Add tags as override attributes in your role, string or key/value
+# When using the Datadog Chef Handler, tags are set on the node with preset prefixes:
+# `env:node.chef_environment`, `role:node.node.run_list.role`, `tag:somecheftag`
 default['datadog']['tags'] = ''
 
 # Collect EC2 tags, set to 'yes' to collect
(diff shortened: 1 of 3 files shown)


miketheman:
@lightpriest Getting back to this feature now. Looks like it can't be merged - there might be a conflict in the datadog.conf somewhere.

Also, is there any tests you might write with ChefSpec to validate the template outcome of passing in the tags attribute?

lightpriest:
@miketheman I added a spec for this, imitating the structure of the other specs. I also tried to add a commit that would fix the merge conflict, but I guess it wasn't enough.

The conflict is in a comment section in `attributes/default.rb` and after my commit I think it should be even simpler to merge. If that's a problem, I'll open another pull-request.

jslusher:
+1 for this!

miketheman:
Closing, as superseded with #296 

Thanks for the contribution!

PR #187, opened 2015-03-04
Remove support of debug mode and add support of custom log levels

remh:
debug_mode has been removed more than 2 years ago in the Agent code:
https://github.com/DataDog/dd-agent/commit/2302ad9479124e96253e5776c5b37e755e6ece67

The proper way to be in "debug mode" is to set the log level accordingly, so this PR is adding this possibilty.


miketheman:
Thanks @remh ! Likely to make it into the next version.

PR #190, opened 2015-03-31
@miketheman configurable process monitoring

jpcallanta:
I've added process.rb recipe as well as the process.yaml.erb as discussed in the thread for pull request #124 and added test-kitchen test.


recipes/process.rb
```diff
@@ -0,0 +1,29 @@
...
+#   },
+#   {
+#     :name => "postgres",
+#     :search_string => ["postgres"],
+#   },
+# ]
+
+datadog_monitor 'process' do
+  instances node['datadog']['process']['instances']
+end
```
miketheman: rubocop might yell at you for lacking a newline here

.kitchen.yml
```diff
@@ -222,3 +222,21 @@ suites:
...
+  run_list:
+    - recipe[datadog::process]
+  attributes:
+    datadog:
+      <<: *DATADOG
+      process:
+        instances:
+          - name: "pidname"
+            exact_match: false
+            git a
```
miketheman: this might not be intended

miketheman:
Nice stuff. I might add a link in the "docs" section to see more over here: https://github.com/DataDog/dd-agent/blob/master/conf.d/process.yaml.example

recipes/process.rb
jpcallanta: Ah, I'll fix

miketheman:
Thanks! I rebased and cleaned up a bit for brevity, and merged in f048bbb62b16f69cfa1fca06228b3a93fc7152f4

PR #197, opened 2015-04-13
[docker] Add recipe and tests

DorianZaccaria:
Add a recipe to create an agent configuration file in order to monitor docker. The attributes available are:
```
datadog: {
  docker: {
    init_config: {
      docker_root: '/',
      socket_timeout: 10
    },
    instances: [
      {
        url: 'unix://var/run/docker.sock',
        new_tag_names: 'false',
        tag_by_command: 'false',
        tags: ['toto', 'tata'],
        include: ['docker_image:ubuntu', 'docker_image:debian'],
        exclude: ['.*'],
        collect_events: 'true',
        collect_container_size: 'false',
        collect_all_metrics: 'false',
        collect_images_stats: 'false',
      }
    ]
  }
}
```


templates/default/docker.yaml.erb
```diff
@@ -0,0 +1,50 @@
...
+  <% if @init_config.key?('docker_root') -%>
+  docker_root: <%= @init_config['docker_root'] %>
+  <% end -%>
+  <% if @init_config.key?('socket_timeout') -%>
+  socket_timeout: <%= @init_config['socket_timeout'] %>
+  <% end -%>
+  <% end -%>
+
+instances:
+  <% @instances.each do |i| -%>
```
miketheman: This is cool, but I think I tried to explain that we should try to use the more flexible method as seen in newer examples like [this one](https://github.com/DataDog/chef-datadog/blob/master/templates/default/http_check.yaml.erb) where the `instances` hash is rendered directly from attributes structure, instead of us needing to change & release whenever an Agent check adds/changes a config flag.

I've been trying to think how this might work for init_config as well, have not figured that out yet.

DorianZaccaria: In that case we don't have any default value for url and new_tag_name. Can we suppose it is users responsibility to always define these attributes?

miketheman: I think we do have many cases where the user is shown an example in the recipe. See here for an example of the http check: https://github.com/DataDog/chef-datadog/blob/master/recipes/http_check.rb#L3-L20

This seems to be a better interface for now, and soon enough we might be able to replace a large amount of identical template files with one standardized file that any monitor would use.
I am considering a breaking change for a 3.0.0 release where we would convert all existing templates and move them to this method. But that's longer term.

PR #201, opened 2015-04-16
[serverspec] Update serverspec template to use new hash syntax

DorianZaccaria:
Use the ruby new hash syntax in serverspec template.


miketheman:
Thanks!

PR #203, opened 2015-04-24
Modified the example of fluentd.rb.

inokappa:
I modified the example of fluentd.rb.

- Fixed from node to node.override
- Add a comma

Please confirm.

Thank you.


miketheman:
Thanks @inokappa !

PR #204, opened 2015-04-24
[supervisord] Add recipe and tests

DorianZaccaria:
Render supervisord configuration for the agent check.


recipes/supervisord.rb
```diff
@@ -0,0 +1,42 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+include_recipe 'datadog::dd-agent'
+
+# Build a data structure with configuration.
```
miketheman: Can you add a link to the actual template in dd-agent? Something like this: https://github.com/DataDog/chef-datadog/blob/master/recipes/jmx.rb#L4

DorianZaccaria: Done.

miketheman:
Thanks!

PR #205, opened 2015-04-30
[doc] Updated Kafka doc for Agent 5.3

elafarge:
Username is now user, just like in dd-agent/conf.d/kafka.yaml.example.


miketheman:
Thanks!

PR #206, opened 2015-05-10
Adding info about using ES integration

nonflet:
Sometimes better to read README,then to read methods how it works :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,17 @@ Usage
     recipe[datadog::dd-handler]
   )
   ```
+  
+  If you need to use, for example, elasticsearch integration , just add :
+  run_list "recipe[datadog::elasticsearch]"
+  default_attributes(
+    'datadog' => {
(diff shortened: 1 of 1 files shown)


miketheman:
@nonflet Thanks for the contrib.

I think that instead of pointing out an example in the README, we describe in more detail on http://docs.datadoghq.com/guides/chef/#customizations - so it might make more sense to link to the docs site, and recommend reading each recipe's inline documentation, as they are likely to be kept more up-to-date than the README.

What do you think?

miketheman:
Also, here's an example of the inline docs for the elasticsearch integration:
https://github.com/DataDog/chef-datadog/blob/7280fe6ba2d8f593b411ca2aba9ffe3f3b31ce94/recipes/elasticsearch.rb#L3-L19

nonflet:
Oooops, my mistake, sorry guys. Missed inline docs.Pls close request.

PR #207, opened 2015-05-20
Enable ability to match with epoch version numbers

coosh:
This fixes the version match if people are running their own repo and have an epoch on the version number. While rare this will keep it compatible with that.

EG:
5.3.2-1
1:5.3.2-1
2:5.3.2-1

All the above will match and find the correct major release version number. 

**NOTE:**
It appears that there are epoch versions on the packages in the ```apt.datadoghq.com``` repository, causing the if statement in its current form to never match correctly.

### DIFF
--- a/recipes/dd-agent.rb
+++ b/recipes/dd-agent.rb
@@ -23,7 +23,7 @@ include_recipe 'datadog::repository' if node['datadog']['installrepo']
 dd_agent_version = node['datadog']['agent_version']
 
 # If version specified and lower than 5.x
-if !dd_agent_version.nil? && dd_agent_version.split('.')[0].to_i < 5
+if !dd_agent_version.nil? && dd_agent_version.match(/(\d{1,})\.\d{1,}\.\d\.*/)[1].to_i < 5
   # Select correct package name based on attribute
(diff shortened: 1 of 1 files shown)


miketheman:
Looks like an interesting enhancement. I think the best solution will be to eventually remove the epoch versions altogether - @remh ?

coosh:
To be strictly honest, removing the epoch version would be a better solution. This is an ugly regex that should go away.

miketheman:
I did some reading, removing the epoch isn't really feasible unless the package name changes - which is unlikely.
We're going to remove the logic entirely at some point, as we no longer expect new installations of Agents below 5.x (which has embedded Python), and that should resolve this.

The condition should only apply to situations where the version is set and it's lower than 5.x - here's the specs: https://github.com/DataDog/chef-datadog/blob/master/spec/dd-agent_spec.rb#L173-L209

@coosh Can you show an example of this that isn't being exercised?

miketheman:
Closing this, while a good addition to the logic, not easily testable, and in our next major revision we will be removing the need to detect for system python versions and the `base` packages.

PR #188, opened 2015-03-20
Pr/5

EasyAsABC123:

PR #164, opened 2014-12-18
added windows agent installation support

EasyAsABC123:
This adds the ability to install the latest (or locally archieved) datadog windows agent.

PR #162, opened 2014-12-18
Support config deployment on Windows

rlaveycal:
See #161 

Current code has unix paths and package deployment preventing operation on Windows. Now split out and parameterised to allow operation.

Working recipes included to deploy iis, sqlserver and wmi integrations on Windows.

PR #210, opened 2015-06-10
Windows agent installation support

### LINKED ISSUE #142: Request: Support Windows agent installation
Hi,

Do you plan to support installation of the Windows agent?

olivielpeau:
Reworked commits from PR #188 and #162 to support agent installation (and upgrade) and deploy checks configuration files on Windows.

Tested successfully on Chef 12. On Chef 11, during an installation run, the agent gets stuck in the 'Stopping' state when requested to stop or restart. I'm still tying to figure out if there's a good workaround.

Many thanks to @EasyAsABC123 and @rlaveycal for their initial PRs.

Closes #142, #188 and #164


miketheman:
Looking pretty awesome. I'll take a deeper review tomorrow and provide some feedback.

Gemfile
```diff
@@ -21,6 +21,6 @@ end
 
 group :integration do
   gem 'kitchen-vagrant'
-  gem 'test-kitchen', '~> 1.3.1'
+  gem 'test-kitchen'
```
miketheman: Why the removal of any version constraint?

attributes/default.rb
```diff
@@ -62,16 +74,17 @@
...
   default['datadog']['install_base'] = Gem::Version.new(node['languages']['python']['version'].gsub(/(\d\.\d\.\d).+/, '\\1')) < Gem::Version.new('2.6.0')
 rescue NoMethodError # nodes['languages']['python'] == nil
-  Chef::Log.warn 'no version of python found, please install Agent version 5.x or higher.'
+  Chef::Log.warn 'no version of python found, please install Agent version 5.x or higher.' unless platform_family?('windows')
 rescue ArgumentError
   Chef::Log.warn "could not parse python version string: #{node['languages']['python']['version']}"
 end
 
 # Agent Version
+# Set to `nil` to install latest (will also upgrade to latest on Windows)
```
miketheman: The comment is a bit misleading - it prompts the user to set it to nil, but it's already nil.

    "Default of `nil` will install latest version. On Windows, this will also upgrade to latest"

metadata.rb
```diff
@@ -21,6 +21,7 @@
 depends          'apt' # We recommend '>= 2.1.0'. See CHANGELOG.md for details
 depends          'chef_handler', '~> 1.1.0'
 depends          'yum'
+depends          'windows'
```
miketheman: The introduction of another cookbook dependency is something that is never "awesome" - are you certain the default providers in Chef 10, 11 and 12 cannot be used?

providers/monitor.rb
```diff
@@ -6,9 +6,11 @@ def whyrun_supported?
 
 action :add do
   Chef::Log.debug "Adding monitoring for #{new_resource.name}"
-  template "/etc/dd-agent/conf.d/#{new_resource.name}.yaml" do
-    owner 'dd-agent'
-    mode 00600
+  template "#{node['datadog']['config_dir']}/conf.d/#{new_resource.name}.yaml" do
+    unless node['platform_family'] == 'windows'
```
miketheman: what ownership and permissions get set if on Windows?

recipes/dd-agent-install-linux.rb
```diff
@@ -0,0 +1,43 @@
+#
+# Cookbook Name:: datadog
+# Recipe:: dd-agent-install-linux
```
miketheman: The recipe name here gets lengthy.

Please use something like `_install-linux.rb` so the final name will look like:

    include_recipe 'datadog::_install-linux' unless platform_family? 'windows'

recipes/dd-agent-install-windows.rb
```diff
@@ -0,0 +1,40 @@
...
+  source node['datadog']['windows_agent_url'] + dd_agent_msi
+  # As of v1.37, the windows cookbook doesn't upgrade the package if a newer version is downloaded
+  # As a workaround uninstall the package first if a new MSI is downloaded
+  notifies :remove, 'windows_package[Datadog Agent]', :immediately
+end
+
+# Install the package
+windows_package 'Datadog Agent' do
+  source temp_file
+  options "APIKEY=\"#{node['datadog']['api_key']}\" HOSTNAME=\"#{node['hostname']}\" TAGS=\"#{node['tags'].join(',')}\""
```
miketheman: You can use Ruby's interpolated string modifier to reduce the amount of quotes that you have to escape, like so:

    options %(APIKEY="#{node['datadog']['api_key']}" HOSTNAME="#{node['hostname']}" TAGS="#{node['tags'].join(',')}")


For reference, `%()` is the same as `%Q()`

recipes/dd-agent.rb
```diff
@@ -17,39 +17,25 @@
...
+# Set the correct config file
+agent_config_file = ::File.join(node['datadog']['config_dir'], 'datadog.conf')
 
 # Make sure the config directory exists
-directory '/etc/dd-agent' do
-  owner 'root'
-  group 'root'
-  mode 0755
+directory "#{node['datadog']['config_dir']}" do
+  unless node['platform_family'] == 'windows'
```
miketheman: Again, if on windows, what happens?

spec/dd-agent_spec.rb
```diff
@@ -12,18 +24,22 @@
...
   end
+end
 
-  it 'enables the datadog-agent service' do
-    expect(@chef_run).to enable_service 'datadog-agent'
+shared_examples_for 'common windows resources' do
+  it_behaves_like 'datadog-agent service'
+
+  it 'ensures the Datadog config directory exists' do
+    expect(@chef_run).to create_directory "#{ENV['ProgramData']}/Datadog"
```
miketheman: Instead of determining and interpolating ENV during the test, set it to something realistic in the block where you construct the Windows object, and then compare with the static value here.
This will eliminate needing to look up an environment variable during testing that is likely not going to be there anyhow.

See [here](https://gist.github.com/brettweavnet/6171363) for an example of how you might stub an environment variable.

miketheman:
@olivielpeau Looking good a bunch of comments.

This supplies updates to chefSpec - what about some test-kitchen tests for this behavior?

metadata.rb
olivielpeau: I wanted to use the default providers at first but ran into a lot of issues with the `windows_package` from core :
- the resource is broken in chef 12.2 and 12.3 (fixed in 12.4), see https://github.com/chef/chef/issues/3316
- using the resource with `action :remove` to uninstall the agent makes the recipe fail if the agent is not installed, instead of simply being up-to-date 
- the resource can upgrade the agent but sometimes breaks it in the upgrade (for instance the upgrade from 5.3.0 to 5.3.1 breaks the agent)
- the original msi needs to be provided to uninstall the agent (whereas the `windows` cookbook handles that nicely by using registry keys to find the uninstaller)

I haven't had any of these issues with the `windows` cookbook.
It seems that the `windows` cookbook is gradually getting merged in chef core, until then I think we should use the cookbook.

olivielpeau:
@miketheman Thanks a lot for your comments! I'll update the PR accordingly and add some test-kitchen tests.

Gemfile
olivielpeau: We actually need `~> 1.4.0` for windows support, I'll change it.

olivielpeau:
@miketheman I've updated the PR, it now has kitchen tests + solves the issues you pointed out.

.kitchen.yml
```diff
@@ -4,6 +4,9 @@ driver:
   customize:
     memory: 1024
 
+provisioner:
+  data_path: test/shared  # directory containing shared helpers for integration tests
```
miketheman: This should not be needed - we already support shared directories via the `test/integration/helpers` subdirectory, so this should also simplify the path handling.

If there's a common spec helper, then let's move all the common `require`s and `set path` into there.

.kitchen.yml
```diff
@@ -27,6 +30,24 @@ platforms:
...
+<% end
+
+windows_platform = 'windows-2012r2'
+
+chef_versions.each do |chef_version|
+%>
+
+- name: <%= windows_platform %>-<%= chef_version %>
+  driver_config:
+    box: mwrock/Windows2012R2  # lightweight windows box
```
miketheman: "light"? `Size: 3.3GB` :grinning:

.kitchen.yml
```diff
@@ -83,6 +104,10 @@ suites:
 - name: datadog_cacti
   run_list:
     - recipe[datadog::cacti]
+  excludes:
+  <% chef_versions.each do |chef_version| %>
+    - <%= windows_platform %>-<%= chef_version %>
+  <% end %>
```
miketheman: Build this array once and use YAML anchors to populate other recipes.

providers/monitor.rb
```diff
@@ -6,9 +6,15 @@ def whyrun_supported?
 
 action :add do
   Chef::Log.debug "Adding monitoring for #{new_resource.name}"
-  template "/etc/dd-agent/conf.d/#{new_resource.name}.yaml" do
-    owner 'dd-agent'
-    mode 00600
+  template "#{node['datadog']['config_dir']}/conf.d/#{new_resource.name}.yaml" do
```
miketheman: If we want to be fully cross-platform path-compliant, we should break elements into strings, and use `File.join`:
```
template File.join(node['datadog']['config_dir'], 'conf.d', "#{new_resource.name}.yaml") do
```

This will leverage the OS' native methods, instead of us relying on `/` working across platforms.

providers/monitor.rb
```diff
@@ -19,9 +25,9 @@ def whyrun_supported?
 end
 
 action :remove do
-  if ::File.exist?("/etc/dd-agent/conf.d/#{new_resource.name}.yaml")
-    Chef::Log.debug "Removing #{new_resource.name} from /etc/dd-agent/conf.d/"
-    file "/etc/dd-agent/conf.d/#{new_resource.name}.yaml" do
+  if ::File.exist?("#{node['datadog']['config_dir']}/conf.d/#{new_resource.name}.yaml")
+    Chef::Log.debug "Removing #{new_resource.name} from #{node['datadog']['config_dir']}/conf.d/"
+    file "#{node['datadog']['config_dir']}/conf.d/#{new_resource.name}.yaml" do
```
miketheman: Same notes with regards to File.join

recipes/_install-linux.rb
```diff
@@ -0,0 +1,43 @@
...
+# limitations under the License.
+#
+
+# Install the Apt/Yum repository if enabled
+include_recipe 'datadog::repository' if node['datadog']['installrepo']
+
+dd_agent_version = node['datadog']['agent_version']
+
+# If version specified and lower than 5.x
+if !dd_agent_version.nil? && dd_agent_version.split('.')[0].to_i < 5
```
miketheman: Let's add a `Chef::Log.deprecation("Support for Agent pre 5.x will be removed in in datadog cookbook version 3.0")` or something like that - so we can also remove anything related to `datadog-agent-base` in the future.

recipes/dd-agent.rb
```diff
@@ -59,10 +49,16 @@
...
 raise "Add a ['datadog']['api_key'] attribute to configure this node's Datadog Agent." if node['datadog'] && node['datadog']['api_key'].nil?
 
-template '/etc/dd-agent/datadog.conf' do
-  owner 'root'
-  group 'root'
-  mode 0644
+template agent_config_file do
+  if node['platform_family'] == 'windows'
+    owner 'Administrators'
+    rights :read, 'Users'
```
miketheman: Should this be permissive? The file contains keys.

recipes/dd-agent.rb
```diff
@@ -59,10 +49,16 @@
...
+    rights :full_control, 'Administrators'
+  else
+    owner 'root'
+    group 'root'
+    mode 0644
```
miketheman: Maybe this is too permissive as well.

miketheman:
@olivielpeau Awesome stuff! Ran through again, added some comments. It's looking better each pass.

recipes/_install-linux.rb
remh: :+1:

.kitchen.yml
olivielpeau: Most boxes are >4GB ; 3.3GB is as "light" as it gets :smiley:

olivielpeau:
Thanks @miketheman for the great review! I've addressed the issues you've found.

I've left the file permissions on Linux unchanged for now: we have a ticket in our backlog to review them on all of our supported installation methods.

jboeshart:
Any idea when this is going to be released? Would love to not have to re-invent the wheel myself with all the good work that's been going on here.

akemner:
+1 we need this

jboeshart:
Talked to someone from DataDog at re:Invent and they said that this would come out with the next major release of the cookbook, which was probably 2-3 months out.

Gemfile
```diff
@@ -12,7 +12,9 @@ end
 
 group :test do
   gem 'berkshelf', '~> 3.1'
-  gem 'chefspec', '~> 4.2'
+  # using this ref so that chefspec doesn't fail when the `rights` attribute is set on the directory resource
+  # FIXME: as soon as the chefspec release following 4.2.0 is out
+  gem 'chefspec', :github => 'sethvargo/chefspec', :ref => '9bb7a6'
```
miketheman: ChefSpec 4.4.0 is out, please update accordingly.

Gemfile
```diff
@@ -21,6 +23,7 @@ end
 
 group :integration do
   gem 'kitchen-vagrant'
-  gem 'test-kitchen', '~> 1.3.1'
+  gem 'test-kitchen', '~> 1.4.0'
   gem 'travis-lint'
+  gem 'serverspec', '~> 2.18.0'
```
miketheman: You shouldn't need to specify a serverspec version here.

providers/monitor.rb
```diff
@@ -19,9 +25,10 @@ def whyrun_supported?
 end
 
 action :remove do
-  if ::File.exist?("/etc/dd-agent/conf.d/#{new_resource.name}.yaml")
-    Chef::Log.debug "Removing #{new_resource.name} from /etc/dd-agent/conf.d/"
-    file "/etc/dd-agent/conf.d/#{new_resource.name}.yaml" do
+  confd_dir = ::File.join(node['datadog']['config_dir'], 'conf.d')
+  if ::File.exist?(::File.join(confd_dir, "#{new_resource.name}.yaml"))
```
miketheman: Clonflict here with recent changes due to why-run mode.

miketheman:
@olivielpeau I merged all outstanding PRs for Agent 2.2.0 and am completing the CHANGELOG now.

I'd love to get this into the release, but there's a couple of outstanding issues here before we can merge cleanly.

This PR is also up to 18 commits already - we need to get it out the door before it grows much more. :grinning: 

Let me know if you have any questions!

olivielpeau:
@miketheman Thanks for the review! I'll work today on the issues you pointed out and try to get things ready by the end of the week.

olivielpeau:
@miketheman I've fixed the issues you found, rebased the PR and re-run the kitchen tests so this is probably ready for a final review :)

miketheman:
@olivielpeau Awesome! I think this is all ready to go, and will merge Monday. We can then start testing it internally, and hopefully find no bugs, then release to the world!

remh:
:tada:

jboeshart:
Much thanks on getting this merged and making it available in a shorter timeline!

PR #212, opened 2015-06-23
Updates to manage LWRP

jmanero-r7:
### Set cookbook attribute for template in manage provider
If the `cookbook` attribute is not defined for a template in an LWRP, the resource will search in the instantiating cookbook, not the defining (`datadog` in this case) cookbook, for its source.

### Restart notifications in manage provider
The manage resource issued restart notifications to the `datadog-agent` service regardless of the `node['datadog']['agent_start']` attribute. This could cause the service to start even though the desired state was stopped.

### DIFF
--- a/providers/monitor.rb
+++ b/providers/monitor.rb
@@ -14,7 +14,7 @@ action :add do
       :instances   => new_resource.instances
     )
     cookbook new_resource.cookbook
-    notifies :restart, 'service[datadog-agent]', :delayed
+    notifies :restart, 'service[datadog-agent]', :delayed if node['datadog']['agent_start']
   end
   new_resource.updated_by_last_action(false)
 end
@@ -24,7 +24,7 @@ action :remove do
     Chef::Log.debug "Removing #{new_resource.name} from /etc/dd-agent/conf.d/"
     file "/etc/dd-agent/conf.d/#{new_resource.name}.yaml" do
       action :delete
-      notifies :restart, 'service[datadog-agent]', :delayed
+      notifies :restart, 'service[datadog-agent]', :delayed if node['datadog']['agent_start']
     end
     new_resource.updated_by_last_action(true)
   end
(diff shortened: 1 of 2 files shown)


miketheman:
@jmanero-r7 That's a good point about the cookbook path. Most of the time, we've directed use of the `datadog::<integrationname>` recipes, which will typically execute the `datadog_monitor` resource at that time.
This was one way to solve the possibility of a node having multiple datadog_monitor invocations for the same integration in differing recipes, overwriting the same config file and restarting the agent multiple times unnecessarily per run.

By building up the `instances` hash on the node object, and including the recipe once, that solved this particular problem.

Can you show an example of using `datadog_monitor` external to the cookbook that does this safely?

I like the `restart` action addition - that's a good one to have. Can you try adding some test cases for it?

jmanero-r7:
Thanks @miketheman. I'll try to get a test added when I have some free time.

> This was one way to solve the possibility of a node having multiple datadog_monitor invocations for the same integration in differing recipes, overwriting the same config file and restarting the agent multiple times unnecessarily per run.

IMO It's really the responsibility of the consumer to handle conflicts if they define a resource. The Chef runtime will generate warnings during the compile phase if multiple resources with the same identity (e.g. `datadog_monitor[cassandra]`) are defined. A previously defined resource can be fetched and modified later in the compile phase if the consumer _really_ wants to. For example (not that this is a good idea, but...):

```ruby
resources('datadog_monitor[cassandra]').instances.push(:host => 'localhost', :port=> 1234)
```

Multiple restarts are a non-issue as the notification is `:delayed`. Since Chef 11, these get queued and de-duplicated, so the `service[datadog-agent]` resource will only have its `:restart` action run once regardless of how many times `notifies :restart, 'service[datadog-agent]'` is triggered.

miketheman:
@jmanero-r7 Thanks!

PR #218, opened 2015-07-10
Change rights on agent configuration files

olivielpeau:
Set dd-agent as owner of /etc/dd-agent and set stricter permissions on datadog.conf

https://github.com/DataDog/dd-agent/pull/1764 applies the same changes to our Unix install scripts.

Mentioned in #210


miketheman:
Thanks!

PR #219, opened 2015-07-23
Add apt-transport-https to the repository install. In preparation for…

darron:
… HTTPS repo.

Without this already installed, when an https repo is added apt fails to update.

With this added it all works properly.

There are no changes needed for the YUM repo.


miketheman:
Thanks @darron !

PR #220, opened 2015-07-29
Add an attribute to enable dd_check tag collection

elafarge:
Verified using test kitchen.


miketheman:
In Agent 5.5.0.

miketheman:
Thanks @elafarge !

PR #223, opened 2015-08-13
add additional option for timeout definition

dominicchan:


miketheman:
Nice. Can you add a doc example to https://github.com/DataDog/chef-datadog/blob/etienne/add-new-es-flags/recipes/elasticsearch.rb#L14-L17 ?

miketheman:
Thanks @dominicchan !

PR #224, opened 2015-09-01
Added remaining options from varnish.yaml.example

hilli:
I was missing the last couple of options that was mentioned in the varnish.yaml.example file - So they are added.


miketheman:
Thanks @hilli !

PR #225, opened 2015-09-01
[yum] Add RPM signature check

elafarge:
FIXME: Use HTTPs URL for both the key and the YUM repository as soon as
we're ready with that.


yannmh:
Succesfully tested with Kitchen on staging YUM repo. :+1:

miketheman:
Our packages are now signed! Thanks @elafarge !

PR #227, opened 2015-09-09
[zk] Make `timeout` attribute optional

### LINKED ISSUE #211: zookeeper check timeout attribute either needs a default or to be optional
In the [config example](https://github.com/DataDog/dd-agent/blob/367819883e3bea937fd178a14a0710ececc3ce7e/conf.d/zk.yaml.example) for the agent's zk check, we see instances.timeout as an optional key:

```
# timeout: 3
```

However, [the Chef template](https://github.com/DataDog/chef-datadog/blob/448e3a97014292eb7d6129a9b1d4240a21ffe0fc/templates/default/zk.yaml.erb) expects the attribute to be set. The result of just setting host and port (as in the example) is a bunch of log errors about timeout not being a string or number:

```
2015-06-17 19:53:58 UTC | ERROR | dd.collector | checks.zk(__init__.py:556) | Check 'zk' instance #0 failed
Traceback (most recent call last):
  File "/opt/datadog-agent/agent/checks/__init__.py", line 547, in run
    self.check(copy.deepcopy(instance))
  File "/opt/datadog-agent/agent/checks.d/zk.py", line 49, in check
    timeout = float(instance.get('timeout', 3.0))
TypeError: float() argument must be a string or a number
```

It would be good if the two agreed with each other on whether it's a mandatory attribute or not.

olivielpeau:
Consistent with the check config in the agent.

Fixes #211

### DIFF
--- a/templates/default/zk.yaml.erb
+++ b/templates/default/zk.yaml.erb
@@ -3,7 +3,9 @@ instances:
   <% @instances.each do |instance| -%>
   - host: <%= instance["host"] %>
     port: <%= instance["port"] %>
+    <% if instance.key?("timeout") -%>
     timeout: <%= instance["timeout"] %>
+    <% end -%>
     <% if instance.key?("tags") -%>
     tags:
       <% instance["tags"].each do |tag| -%>


miketheman:
Thanks @olivielpeau !

PR #229, opened 2015-09-10
Fix `updated_by_last_action` value of `monitor` provider

olivielpeau:
Replaces #168 (see discussion there for more context on this issue)

`use_inline_resources` is the way to go to ensure that `updated_by_last_action` is correct.

It makes the provider actions run in their own run context, so we need to redefine the `datadog-agent` service in each action.

The behavior of the provider is preserved, except that the `datadog-agent` service gets restarted every time a config file is updated (instead of only being restarted once) when multiple files are updated in the same run.

### DIFF
--- a/providers/monitor.rb
+++ b/providers/monitor.rb
@@ -1,5 +1,8 @@
 # Creates the proper yaml file in /etc/dd-agent/conf.d/
 
+# Defined since Chef 11
+use_inline_resources if defined?(use_inline_resources)
+
 def whyrun_supported?
   true
 end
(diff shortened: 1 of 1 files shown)


miketheman:
@olivielpeau I'm not certain I follow. I think the desire is to have datadog-agent restart once, at the end of the run, if there are any services notifying it to do so - this appears to define extra resources with the same name (which causes the infamous [resource cloning](https://github.com/chef/chef-rfc/issues/76)), and I'm not sure that's the correct behavior, since we can have multiple restarts per addition to the file.

I'm also not sure about the removal of the `updated_by_last_action` - that's something we'd likely have to test - can you conceive of a way to test that?

olivielpeau:
@miketheman I agree that restarting the service multiple times per run is a big issue...

Given that the provider's actions are executed in their own run contexts, the only way that I see to notify the service in the main run context would be to remove the notifications in the provider and add:
```
notifies :restart, 'service[datadog-agent]', :delayed
```
to every `datadog_monitor` resource block. But that changes the scope of what the resource does.
Unless I'm missing something.

I'm not sure what we could test exactly on the `updated_by_last_action`. Basically its current value doesn't make much sense (for instance it's always `false` for the `add` action, even when the config file has been changed). But I'm not sure I understand what you mean by testing its removal.

We could definitely test the whole provider with chefspec though.

miketheman:
Thanks @olivielpeau !

Issue #118, opened 2014-04-28
Support 'why-run' mode


miketheman:

olivielpeau:
Tested on Chef 12.2.1, the why-run execution seems to work fine (no changes to the state and no event/metrics emitted).

The only case I've found that doesn't work is the `dd-handler` recipe when the `chef-handler-datadog` gem is not installed. The run fails when `'chef/handler/datadog'` is required [here](https://github.com/DataDog/chef-datadog/blob/7e81689ab6c972337409a4707ff46ab44bbe785e/recipes/dd-handler.rb#L29), with the following error:

```
  ================================================================================
  Recipe Compile Error in /home/vagrant/cookbooks/datadog/recipes/dd-handler.rb
  ================================================================================

  LoadError
  ---------
  cannot load such file -- chef/handler/datadog

  Cookbook Trace:
  ---------------
    /home/vagrant/cookbooks/datadog/recipes/dd-handler.rb:29:in `from_file'
```

(obviously because in why-run mode the `chef_gem` resource doesn't actually install the gem that's required)

But from what I understand this case is unavoidable.

miketheman:
@olivielpeau Thanks for the investigation. Regarding the final case, this was [reported a while back](https://tickets.opscode.com/browse/CHEF-4391), and claimed to have been resolved. Might be worthwhile looking into it a bit further?

olivielpeau:
@miketheman I think the issue you're mentioning is that the `chef_gem` resource itself breaks why-run mode, whereas here it's the subsequent `require` on `chef/handler/datadog` that's making the run fail when the file doesn't exist.

I'll look more into this issue (in particular I'll look at how chef_handler loads the handler's class and if that can help us solve this).

miketheman:
@olivielpeau You're right - I linked the wrong issue.

Thinking about this further, there might be a better way to wrap the require that behaves differently based on whether or not we're in why-run mode, by inspecting the value of `Chef::Config[:why_run]`.

Something maybe like:
```ruby
if Chef::Config[:why_run]
  Chef::Log.warn('Running in why-run mode, skipping dd-handler')
  return
end
```

at the top of the recipe.

olivielpeau:
@miketheman I like the idea of skipping `dd-handler` in why-run mode as I can't see any clean workaround to execute it in this mode.

[...]

PR #231, opened 2015-09-16
[why-run] Skip dd-handler recipe in why-run mode

olivielpeau:
Running in why-run mode fails when the chef-handler-datadog gem is not
installed because of the `require 'chef/handler/datadog'` line, so we
skip the entire recipe in why-run mode.

Upgrading chef_handler to v1.2 (Chef>=12)  will allow us to remove that
`require` and unskip the recipe in why-run mode.

See #118

### DIFF
--- a/recipes/dd-handler.rb
+++ b/recipes/dd-handler.rb
@@ -17,6 +17,14 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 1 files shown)


miketheman:
Thanks @olivielpeau !

PR #232, opened 2015-09-17
Add missing options for rabbitmq collector

mattrobenolt:
Adds the following: nodes, nodes_regexes, queues, queues_regexes, and vhosts

### DIFF
--- a/templates/default/rabbitmq.yaml.erb
+++ b/templates/default/rabbitmq.yaml.erb
@@ -6,10 +6,40 @@ instances:
   - rabbitmq_api_url: <%= i['api_url'] %>
     rabbitmq_user: <%= i['user'] || 'guest' %>
     rabbitmq_pass: <%= i['pass'] || 'guest' %>
-    tags:
   <% if i.key?('tags') -%>
-    <% i['tags'].each do |t| -%>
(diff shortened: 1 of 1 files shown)


templates/default/rabbitmq.yaml.erb
```diff
@@ -6,10 +6,40 @@ instances:
   - rabbitmq_api_url: <%= i['api_url'] %>
     rabbitmq_user: <%= i['user'] || 'guest' %>
     rabbitmq_pass: <%= i['pass'] || 'guest' %>
-    tags:
```
mattrobenolt: This also makes `tags` get rendered out optionally.

miketheman:
Hi @mattrobenolt!

Thanks for taking the time. We're trying to get into a model where the templates are simpler, and can be rendered from _any_ Hash, allowing us to not have to update & release after any change in the Agent config.

This leads to simpler template code, and becomes a bit easier to maintain.

If you'd like to take the time to revisit this, we'd appreciate it!

fhanson:
@miketheman is there a rewrite pending for converting to use any hash? If not this PR fixes a major problem for folks monitoring large rabbitmq setups.

PR #233, opened 2015-09-22
Add notify on failure of chef-client run

jonathanmorley:
The dd-handler does not currently support notifications when chef fails.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -81,7 +81,7 @@ default['datadog']['chef_handler_version'] = nil
 default['datadog']['chef_handler_enable'] = true
 
 # List to notify in case of chef failures
-default['datadog']['handler']['notify_on_failure'] = []
+default['datadog']['handler']['notify_on_failure'] = nil
 
 # Log level. Should be a valid python log level https://docs.python.org/2/library/logging.html#logging-levels
 default['datadog']['log_level'] = 'INFO'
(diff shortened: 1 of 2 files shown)


recipes/dd-handler.rb
```diff
@@ -34,7 +34,8 @@
   arguments [
     :api_key => node['datadog']['api_key'],
     :application_key => node['datadog']['application_key'],
-    :use_ec2_instance_id => node['datadog']['use_ec2_instance_id']
+    :use_ec2_instance_id => node['datadog']['use_ec2_instance_id'],
+    :notify_on_failure => node['datadog']['handler']['notify_on_failure']
```
miketheman: nitpick: lists should be alpha-sorted, unless ordering makes a difference.

attributes/default.rb
```diff
@@ -80,6 +80,9 @@
 # Enable the Chef handler to report to datadog
 default['datadog']['chef_handler_enable'] = true
 
+# List to notify in case of chef failures
+default['datadog']['handler']['notify_on_failure'] = []
```
miketheman: Should the default be an array? I think it should be `nil` and activated with `true`

miketheman: Sorry, not `true`, rather an array of handles.  But `nil` should be the default.

miketheman:
@jonathanmorley Thanks for working through this - I now realize that we had released (and blogged about) Event Alerts being defined inside Datadog's interface. See https://www.datadoghq.com/blog/event-alerts-another-way-to-trigger-notifications/

I realize this PR was developed before Event Alerts were widely discussed, how would you feel about leaving this aside for now?

jonathanmorley:
:+1:

miketheman:
Closing this PR for now, thanks @jonathanmorley . If Event Alerting doesn't pan out, we can re-discuss.

PR #236, opened 2015-09-26
Fixing errors created by process template using to_json & to_yaml.

zshenker:
This sets up a defined template for process.


miketheman:
@zshenker Thanks for taking the time to do this, but we'e trying to get away from the "specify each possible config option as an ERB tag" and use the more open-ended JSON style, keeping the template lightweight.

If there a particular bug that you can show that requires this change?

PR #241, opened 2015-10-16
Let the agent set the default hostname

remh:
The hostname logic in the agent is more reliable

Defaulting to the node name can create some issues as the chef node
name can be based on the ip of the machine that could get recycled.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -100,7 +100,7 @@ default['datadog']['check_freq'] = 15
 
 # Specify agent hostname
 # More information available here: http://docs.datadoghq.com/hostnames/#agent
-default['datadog']['hostname'] = node.name
+default['datadog']['hostname'] = nil
 
 # If running on ec2, if true, use the instance-id as the host identifier
 # rather than the hostname for chef-handler.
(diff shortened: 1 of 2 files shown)


miketheman:
Will the agent detect a chef node name?

remh:
It will use the instanceid in case of EC2/Google Cloud, then default on the machine hostname.

remh:
cc: @DBLaw

miketheman:
@remh Thanks, I think I get it. I think this might have to wait until 3.0.0, as the change removes the default setting of an attribute to `nil`, something that was defined in the CVP as a major change.
Ref: http://chef-community.github.io/cvp/

Considering that anyone wanting to "remove" this can override the attribute with their choice, using `node.fqdn`, `node.hostname` or any other value, or even apply a `node.rm('datadog', 'hostname')` ([ref](http://jtimberman.housepub.org/blog/2014/12/24/quicktip-delete-attributes/)) as a workaround, I think we should leave this until the next Major release.

AliDatadog:
Closed for inactivity

PR #242, opened 2015-10-19
Use non-deprecated python datadog library

jonathanmorley:
The current python package (https://github.com/DataDog/dogstatsd-python) is deprecated in favour of the 'datadog' package (https://github.com/DataDog/datadogpy)

### DIFF
--- a/recipes/dogstatsd-python.rb
+++ b/recipes/dogstatsd-python.rb
@@ -17,4 +17,4 @@
 # limitations under the License.
 #
 
-easy_install_package 'dogstatsd-python'
+easy_install_package 'datadog'


miketheman:
@jonathanmorley Thanks for changing this! I think it'll have to wait until the next major release of the cookbook, as it pulls in a different library than before.

miketheman:
@jonathanmorley I wonder if it would make more sense to add a second recipe that does the new thing, and add a deprecated message to this one instead of swapping it under the hood.

olivielpeau:
If this change is still needed, it should be done by adding a new recipe that installs the newer `datadog` python package. Closing the PR, feel free to reopen it if you can make the change

PR #244, opened 2015-10-21
Fix tags template in kafka check

LeoCavaille:
`@tags` is never defined so if a `tags` key is present in the instance,
chef will throw:

```
undefined method `each' for nil:NilClass
```


miketheman:
Thanks!

PR #245, opened 2015-10-23
Example doesn't work when it is rendered

tejom:
When this example is copied the line `'port' => '7199' `will be rendered as '7199'  in the template. Data dog complains about the port value being a non integer and the stats collection won't work. Small change but caused a headache.


remh:
Looks good to me, thanks!

PR #247, opened 2015-10-24
add guard for old package removal

dwradcliffe:
This will skip the removal of the `datadog-agent-base` package if it is not installed.
Due to how chef [checks to see if a package is installed](https://github.com/chef/chef/blob/8bf1304da739d4be94edb101ad9e46c96b1d4ccd/lib/chef/provider/package/apt.rb#L79-L97), and the fact that this package is a virtual package, chef always tries to uninstall the package even if it is not installed.

### DIFF
--- a/recipes/_install-linux.rb
+++ b/recipes/_install-linux.rb
@@ -35,6 +35,8 @@ else
   # default behavior, remove the `base` package as it is no longer needed
   package 'datadog-agent-base' do
     action :remove
+    not_if 'rpm -q datadog-agent-base' if %w(rhel fedora).include?(node['platform_family'])
+    not_if 'apt-cache policy datadog-agent-base | grep "Installed: (none)"' if node['platform_family'] == 'debian'
   end
   # Install the regular package
   package 'datadog-agent' do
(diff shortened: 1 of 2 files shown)


miketheman:
Hi there!
Thanks for taking the time to report this.
We will be removing the package removal on the next major version of the
cookbook, as it served a transitional purpose for the pre-version-5 Agent
installs.
The guard added is apt-specific, and likely would not work on rpm-based
installs.

Is there a better guard that you might come up with?

dwradcliffe:
Added guards for both platform families. Also fixed the specs so they should pass now.

miketheman:
@dwradcliffe Thanks! The specs aren't being run due to style failures. These can be run locally when running `rake` in the root of the repo.

Honestly, this seems like a failing of Chef, if it's not doing the right thing in the case of a virtual package.

dwradcliffe:
I agree, but after looking at the code I didn't see an easy path forward that doesn't break existing functionality on that end. I was hoping to take a simpler short cut. How long until this resource is removed entirely?

dwradcliffe:
Style tests fixed

miketheman:
Thanks @dwradcliffe!

PR #248, opened 2015-10-24
Add source & issues URLs for Supermarket

jeffbyrnes:
Sets the URLs for the Supermarket in code, allowing visitors arriving at the Supermarket page to find their way easily to the GitHub project.


miketheman:
Awesome, thanks! Will merge Monday prior to release.

PR #249, opened 2015-10-26
[docker-daemon] Add recipe and tests

kurochan:
Add a recipe to create an agent configuration file in order to monitor docker with "docker_daemon" check.
Because "docker" check is deprecated.
 The attributes available following:

```yaml
datadog: {
  docker: {
    init_config: {
      docker_root: '/',
      socket_timeout: 10,
      tls: false,
      tls_client_cert: '/path/to/client-cert.pem',
      tls_client_key: '/path/to/client-key.pem',
      tls_cacert: '/path/to/ca.pem',
      tls_verify: true
    },
    instances: [
      {
        url: 'unix://var/run/docker.sock',
        tags: [
          'toto',
          'tata'
        ],
        include: [
          'docker_image:ubuntu',
          'docker_image:debian'
        ],
        exclude: [
          '.*'
        ],
        performance_tags: [
          'container_name',
          'image_name',
          'image_tag',
          'docker_image'
        ],
        container_tags: [
          'image_name',
          'image_tag',
          'docker_image'
        ],
        collect_labels_as_tags: [
          'com.docker.compose.service',
          'com.docker.compose.project'
        ],
        ecs_tags: true,
        collect_events: true,
        collect_container_size: false,
        collect_image_size: false,
        collect_images_stats: false
      }
    ]
  }
}
```


templates/default/docker_daemon.yaml.erb
```diff
@@ -0,0 +1,28 @@
+# Generated by Chef, local modifications will be overwritten
+
+init_config:
```
miketheman: @kurochan Thanks! Could you possibly try to replace this section with a similar one to the `instances` one? This retains flexibility in the Chef template if/when the Agent config changes.

Something like:
```ruby
<%= JSON.parse(({ 'init_config' => @ init_config }).to_json).to_yaml %>
```

miketheman:
Nice stuff, thanks!

I added a note in the template that should be addressed.

Also, you may want to add a `spec` test for faster feedback - this is something I have started to do with recent changes. See https://github.com/DataDog/chef-datadog/blob/master/spec/integrations/redis_spec.rb for an example.

Running `rspec spec/integrations/redis_spec.rb` is _much_ faster than running a complete kitchen test.

kurochan:
@miketheman Thank you for reviewing.
I fixed spec and template. Is this OK?

miketheman:
@kurochan This looks good, thank you!

PR #252, opened 2015-11-03
Update windows support to be better

brentm5:
Things changed
- Made the msi file install perMachine instead of per user
- Do not allow the program to restart (Need more feedback on this)
- Added tests to verify the program is installed perMachine
- Add checksum to verify file downloaded is who it says it is
- Removed chef 11.X from test kitchen for windows because its not supported

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -30,8 +30,11 @@ platforms:
 <% end
 
 windows_platform = 'windows-2012r2'
+windows_chef_versions = %w(
+  12.2.1
+)
 
-chef_versions.each do |chef_version|
+windows_chef_versions.each do |chef_version|
 %>
 
 - name: <%= windows_platform %>-<%= chef_version %>
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,7 +8,7 @@ AllCops:
 
 # We like longer lines
 Metrics/LineLength:
-  Max: 199
+  Max: 160
 
 # TODO: add encoding headers
 Style/Encoding:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,7 +8,7 @@ AllCops:
 
 # We like longer lines
 Metrics/LineLength:
-  Max: 160
+  Max: 199
 
 # TODO: add encoding headers
 Style/Encoding:
--- a/recipes/_install-windows.rb
+++ b/recipes/_install-windows.rb
@@ -24,9 +24,9 @@ dd_agent_msi = dd_agent_version ? "ddagent-cli-#{dd_agent_version}.msi" : 'ddage
 temp_file = ::File.join(Chef::Config[:file_cache_path], 'ddagent-cli.msi')
 
 # Download the installer to a temp location
-remote_file 'MSI installer' do
-  path temp_file
+remote_file temp_file do
   source node['datadog']['windows_agent_url'] + dd_agent_msi
+  checksum node['datadog']['windows_agent_checksum'] if node['datadog']['windows_agent_checksum']
   # As of v1.37, the windows cookbook doesn't upgrade the package if a newer version is downloaded
   # As a workaround uninstall the package first if a new MSI is downloaded
   notifies :remove, 'windows_package[Datadog Agent]', :immediately
@@ -35,6 +35,8 @@ end
 # Install the package
 windows_package 'Datadog Agent' do
   source temp_file
-  options %(APIKEY="#{node['datadog']['api_key']}" HOSTNAME="#{node['hostname']}" TAGS="#{node['tags'].join(',')}")
+  installer_type :msi
+  options '/norestart ALLUSERS=1'
   action :install
+  success_codes [0, 3010]
 end
--- a/recipes/_install-windows.rb
+++ b/recipes/_install-windows.rb
@@ -33,7 +33,7 @@ remote_file temp_file do
 end
 
 # Install the package
-windows_package 'Datadog Agent' do
+windows_package 'Datadog Agent' do # ~FC009
   source temp_file
   installer_type :msi
   options '/norestart ALLUSERS=1'
--- a/test/integration/dd-agent/serverspec/dd-agent_spec.rb
+++ b/test/integration/dd-agent/serverspec/dd-agent_spec.rb
@@ -8,11 +8,6 @@ describe service(@agent_service_name) do
   it { should be_running }
 end
 
-# verifies that the agent is installed per-machine for windows
-describe windows_registry_key('HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Installer\UserData\S-1-5-18\Products\776BC1A520EDC164EBF87A0875775D9C'), :if => os[:family] == 'windows' do
-  it { should exist }
-end
-
 describe command('/etc/init.d/datadog-agent info'), :if => os[:family] != 'windows' do
   its(:exit_status) { should eq 0 }
   its(:stdout) { should contain 'OK' }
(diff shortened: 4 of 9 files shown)


recipes/_install-windows.rb
```diff
@@ -35,6 +35,8 @@
 # Install the package
 windows_package 'Datadog Agent' do
   source temp_file
-  options %(APIKEY="#{node['datadog']['api_key']}" HOSTNAME="#{node['hostname']}" TAGS="#{node['tags'].join(',')}")
```
bigbam505: This really isn't needed as its entirely managed by the template unless I am missing something?

recipes/_install-windows.rb
```diff
@@ -35,6 +35,8 @@
...
+  installer_type :msi
+  options '/norestart ALLUSERS=1'
```
bigbam505: Makes sure that the service does not restart and that it is installed globally for all users.  This will cause a bunch of issues down the road since the dd-agent msi does not install on a perMachine bases as illustrated by the following issue https://github.com/DataDog/dd-agent/issues/1689

miketheman:
I'll defer initial review to @olivielpeau .

bigbam505:
Also just as a little FYI without some of the changes in here a bunch of people are going to have a bad time installing datadog with chef.  The main thing is the Per user install.  At first it worked great but then it ended up creating so many problems for our infrastructure later down the road!

.kitchen.yml
```diff
@@ -30,8 +30,11 @@ platforms:
 <% end
 
 windows_platform = 'windows-2012r2'
+windows_chef_versions = %w(
+  12.2.1
+)
 
-chef_versions.each do |chef_version|
+windows_chef_versions.each do |chef_version|
```
olivielpeau: Thanks for spotting this :)

recipes/_install-windows.rb
```diff
@@ -24,17 +24,19 @@
 temp_file = ::File.join(Chef::Config[:file_cache_path], 'ddagent-cli.msi')
 
 # Download the installer to a temp location
-remote_file 'MSI installer' do
-  path temp_file
+remote_file temp_file do
   source node['datadog']['windows_agent_url'] + dd_agent_msi
+  checksum node['datadog']['windows_agent_checksum'] if node['datadog']['windows_agent_checksum']
```
olivielpeau: The `checksum` is used to prevent Chef from re-downloading the file (based on Chef's docs). And to prevent re-downloading the file this resource already uses the etag given in the response's headers. Do you have a case in mind where providing the checksum here gives better results?

recipes/_install-windows.rb
```diff
@@ -24,17 +24,19 @@
...
   notifies :remove, 'windows_package[Datadog Agent]', :immediately
 end
 
 # Install the package
-windows_package 'Datadog Agent' do
+windows_package 'Datadog Agent' do # ~FC009
   source temp_file
-  options %(APIKEY="#{node['datadog']['api_key']}" HOSTNAME="#{node['hostname']}" TAGS="#{node['tags'].join(',')}")
+  installer_type :msi
+  options '/norestart ALLUSERS=1'
```
olivielpeau: Have you seen cases where the installation asks for a restart?

bigbam505: Yes we have in the past

recipes/_install-windows.rb
bigbam505: One positive to this is also for security, you make sure the file you are downloading is the correct file.  For most of our windows downloads we usually specify the version / checksum which allows for locking down a version however this is more painful when new releases ships from a cookbook management level.  Since it is optional it seemed like a good option to have available for people that want to use.

test/integration/dd-agent/serverspec/dd-agent_spec.rb
```diff
@@ -8,6 +8,11 @@
   it { should be_running }
 end
 
+# verifies that the agent is installed per-machine for windows
+describe windows_registry_key('HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Installer\UserData\S-1-5-18\Products\776BC1A520EDC164EBF87A0875775D9C'), :if => os[:family] == 'windows' do
```
olivielpeau: The registry key is different on my test machine so the test fails. Not sure how the key is generated so I'm not sure how we could test this reliably.

recipes/_install-windows.rb
olivielpeau: Thanks for the detailed explanation! Actually you're right, the resource _does_ check that the downloaded file matches the provided checksum, and fails if not (although the docs only mention that the checksum is used to decide whether to re-download the existing file or not).

Could you add a comment on how to use the attribute in `attributes/default.rb`?

test/integration/dd-agent/serverspec/dd-agent_spec.rb
bigbam505: I was afraid of this.  I ran it twice with two kitchen converges and it worked the same way both times.  But if its none deterministic I will remove it.

recipes/_install-windows.rb
bigbam505: Added a comment

.rubocop.yml
```diff
@@ -8,7 +8,7 @@ AllCops:
 
 # We like longer lines
 Metrics/LineLength:
-  Max: 160
+  Max: 199
```
olivielpeau: You don't need to change this value anymore, I think you changed it for the test that is now removed ;)

olivielpeau:
Just added one last comment, apart from that it looks good to me!

@miketheman could you have a look at this PR?

recipes/_install-windows.rb
```diff
@@ -24,17 +24,19 @@
...
   source node['datadog']['windows_agent_url'] + dd_agent_msi
+  checksum node['datadog']['windows_agent_checksum'] if node['datadog']['windows_agent_checksum']
   # As of v1.37, the windows cookbook doesn't upgrade the package if a newer version is downloaded
   # As a workaround uninstall the package first if a new MSI is downloaded
   notifies :remove, 'windows_package[Datadog Agent]', :immediately
 end
 
 # Install the package
-windows_package 'Datadog Agent' do
+windows_package 'Datadog Agent' do # ~FC009
```
miketheman: @bigbam505 can you please clarify what resource attribute is not being recognized here?

bigbam505: Its the windows_package resource 

https://travis-ci.org/DataDog/chef-datadog/jobs/89124906#L222

miketheman:
I'm going to look a little further at the resource problem before merging, but otherwise nice job! Thanks!

bigbam505:
@miketheman So here is the issue I believe, it has to do with using the windows_package from chef or from the windows package

They have two different specs for the return codes

windows cookbook resource
https://github.com/chef-cookbooks/windows/blob/4af5c9184f277ce62fe775693a840367265a65ec/libraries/windows_package.rb#L202

Chef resource
https://github.com/chef/chef/blob/master/lib/chef/resource/windows_package.rb#L44

Because of this inconsistency the `success_codes` property is invalid for this.  If you include the windows cookbook it will take precedence over the included resource in chef so you need to use `success_codes` instead of `returns`

bigbam505:
poke... any traction on this?

bigbam505:
Going to just close this since no one cares about windows.

olivielpeau:
Hi @bigbam505, sorry for the late reply, we're definitely interested in including this in a future release of the cookbook.

One thing I'm concerned about is that changing the installation method to a `per-machine` installation could be a BC-breaking change for existing users of the cookbook, I want to give this some testing to make sure that it's not the case.

If there's no BC break we'll definitely ship this PR with the next minor release. I'll keep you updated. And again, sorry for the long delay in giving you feedback on your PR.

olivielpeau:
Merged through #334 and released in `2.5.0`, thanks again @brentm5!

Closing

brentm5:
Thanks @olivielpeau

PR #253, opened 2015-11-04
Add PHP-FPM recipe

jridgewell:


miketheman:
Hi @jridgewell !

Thanks for taking the time to submit this - we appreciate it! I don't know if you've been following, we've been trying to simplify the template structure to a more open-ended one, where the attribute configuration drives the content of file.
See https://github.com/DataDog/chef-datadog/blob/master/templates/default/jmx.yaml.erb for an example.

This allows the template to be as flexible as the data you put into it, so the cookbook doesn't need an update whenever a flag changes.

Does that make sense?

jridgewell:
Updated.

> This allows the template to be as flexible as the data you put into it, so the cookbook doesn't need an update whenever a flag changes.

With these being so generic, why use multiple template files? You could do something like

```erb
# template.yaml.erb
<%= JSON.parse(({ 'instances' => @instances, 'init_config' => @init_config }).to_json).to_yaml %>
```

```ruby
action :add do
  template new_resource.name do
    source  'template.yaml'
    path ::File.join(node['datadog']['config_dir'], 'conf.d', "#{new_resource.name}.yaml")
    if node['platform_family'] == 'windows'
      owner 'Administrators'
      rights :full_control, 'Administrators'
      inherits false
    else
      owner 'dd-agent'
      mode 00600
    end
    variables(
      :init_config => new_resource.init_config,
      :instances   => new_resource.instances
    )
    cookbook new_resource.cookbook
    notifies :restart, 'service[datadog-agent]', :delayed if node['datadog']['agent_start']
  end

  service 'datadog-agent' do
    service_name node['datadog']['agent_name']
  end
end
```

miketheman:
@jridgewell Thanks! Yes, that one direction we're considering going with the templates, especially as more and more get ported to this format.

There's still many that have a semi-rigid init_config, so that's something we'd have to tackle in the Agent, or here.

PR #257, opened 2015-11-13
Update Kafka configuration to 0.8.2 format

dougbarth:
This changes the Chef template to render a file that closely aligns with the [example configuration](https://github.com/DataDog/dd-agent/pull/2079) provided with the dd-agent code. More details on the changes in the configuration file can be found the dd-agent repository.

With this change, the rendered file will only correctly collect metrics on brokers that are on 0.8.2 or higher.

### DIFF
--- a/spec/integrations/kafka_spec.rb
+++ b/spec/integrations/kafka_spec.rb
@@ -1,0 +1,93 @@
+describe 'datadog::kafka' do
+  cached(:chef_run) do
+    ChefSpec::SoloRunner.new(step_into: ['datadog_monitor']) do |node|
+      node.automatic['languages'] = { 'python' => { 'version' => '2.7.2' } }
+
+      node.set['datadog'] = {
+        'api_key' => 'someapikey',
+        'kafka' => {
+          'instances' => [
+            {
+              'host' => 'localhost',
+              'port' => 1234
+            }
+          ]
+        }
+      }
+    end.converge(described_recipe)
+  end
+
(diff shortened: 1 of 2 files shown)


miketheman:
Thanks! Before we merge this in, I'd like to see how the dd-agent discussion goes.
I also don't see any updates to tests - we've added a more convenient method with ChefSpec - see an [example here](https://github.com/DataDog/chef-datadog/blob/master/spec/integrations/solr_spec.rb).

yannmh:
I am in favor of versioning the template here, like we'll be doing for Cassandra, c.f. https://github.com/DataDog/chef-datadog/pull/263

Users would, then, be able to pick the right template according to Kafka version.

felka:
@dougbarth Thanks a lot for working kafka.yaml! 
I just added erb to my wrapper kafka cookbook and run datadog monitor resource in case kafka version is more than 0.8.2. Resource takes template from the wrapper cookbook. For older versions of kafka it fallbacks to run recipe datadog::kafka which deploys older JMX Mbean names.

olivielpeau:
I agree with @yannmh, I think having a versioned template (similar to what #263 for the cassandra recipe) would be better.

Feel free to implement a versioned template. I'm also adding this PR to our triage milestone so that we make sure to implement the versioned template, if you can't.

olivielpeau:
Solved by #340. Closing, thanks @dougbarth!

Issue #260, opened 2015-12-07
Windows IIS config not setting Sites


cobusbernard:
When you use the IIS recipe, the sites list is not set. This can been seen here: https://github.com/DataDog/chef-datadog/blob/master/templates/default/iis.yaml.erb

Steps to reproduce:
 * Add attribute for Sites via: `default['datadog']['iis']['instances'] = [ { "sites" => ["Default Web Site" ] } ]`
 * After chef-client runs, the iis.yaml will not have a "sites" node.

PR #261, opened 2015-12-07
This will add the sites config to the yaml file generated for iis

cobusbernard:
This will add the 'sites' values to the generated iis.yaml for the DataDog agent on Windows. #260

### DIFF
--- a/recipes/iis.rb
+++ b/recipes/iis.rb
@@ -8,7 +8,8 @@ include_recipe 'datadog::dd-agent'
 # node.datadog.iis.instances = [
 #                               {
 #                                 "host" => "localhost",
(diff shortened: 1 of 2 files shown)


miketheman:
Hi @cobusbernard !

Thanks for the patches - I think the Marathon addition deserves its own PR.

Also, we don't accept patches containing modification to the `version` attribute in metadata.rb - please keep the patch to the code/tests/docs changes only.

cobusbernard:
@miketheman Done. Will create another PR for the Marathon part.

cobusbernard:
Hi @miketheman, stripped out the Marathon part as a separate PR and updated this PR as well.

PR #264, opened 2015-12-14
update kafka template to match kafka current sample yaml

jhohertz:
The datadog::kafka recipe will not work with current kafka versions, and does not reflect the sample YAML shipped in the main product.

This patch brings the kafka.yaml.erb template to match the current kafka.yaml.sample file.


jeffbyrnes:
I think this is a dupe of #257, and is a problem I ran into myself.

olivielpeau:
Fixed in #340, closing

PR #265, opened 2015-12-15
[dd-agent][windows] install per user → per machine

yannmh:
Set `ALLUSERS=1` to force install Datadog Agent on Windows per machine
rather than per user.

This avoid having multiple versions of Datadog Agent installed or
overriding whenever chef-client is run by different users.

This causes no backward incompatibility as Datadog Agent is always
removed before a new install.

@olivielpeau can you have a look please ?

### DIFF
--- a/recipes/_install-windows.rb
+++ b/recipes/_install-windows.rb
@@ -35,6 +35,6 @@ end
 # Install the package
 windows_package 'Datadog Agent' do
   source temp_file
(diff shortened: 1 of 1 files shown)


olivielpeau:
#252 implements this change and is close to being merged, is there a specific reason to open a separate PR?

yannmh:
Closing in favor of #252 :sweat_smile:

PR #268, opened 2016-01-14
Add support for collect_instance_metadata and collect_security_groups config options.

moneill:
This adds support for the `collect_instance_metadata` and `collect_security_groups` configuration options supported by `datadog.conf`. See https://github.com/DataDog/dd-agent/blob/master/datadog.conf.example for more information.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -42,6 +42,12 @@ default['datadog']['create_dd_check_tags'] = nil
 # Collect EC2 tags, set to 'yes' to collect
 default['datadog']['collect_ec2_tags'] = nil
 
+# Collect instance metadata, set to 'yes' to collect
+default['datadog']['collect_instance_metadata'] = nil
+
+# Incorporate security-groups into tags collected from AWS EC2, set to 'yes' to collect
+default['datadog']['collect_security_groups'] = nil
+
 # Autorestart agent
 default['datadog']['autorestart'] = false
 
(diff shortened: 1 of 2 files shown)


moneill:
Oops, apologies for the open-close-open nonsense! Spot checking on some of our instances this appears to be setting the properties correctly in `datadog.conf`, so hopefully some other folks will find this useful!

miketheman:
Thanks! Would sure be nice to see some tests added to cover this behavior.

moneill:
@miketheman sure thing, I was running off of this as an example: https://github.com/DataDog/chef-datadog/pull/159/files. Can you point me to something similar with tests and I'll get those added?

miketheman:
@moneill That's fair. I just looked, and it turns out we don't have any spec tests that encompass the base config file. Sigh.

miketheman:
@remh There is an open question from [this commit](https://github.com/DataDog/dd-agent/commit/ee5d2259a07592c375437813338fc9aef64c6130).

degemer:
Hello @moneill !
Sorry for the late response, it is now (since 2.7.0) possible to use any configuration option in `datadog.conf`: https://github.com/DataDog/chef-datadog/pull/375. Since these options are not often used (`collect_instance_metadata` is by default true), I don't think it would make sense to add them to our recipe (now that there is another way to use them). I'm going to go ahead and close this PR, let me know if you think it should be available by default.

moneill:
Sounds great @degemer, thanks for following up!

PR #272, opened 2016-02-04
expose histogram settings as node attributes

SupermanScott:
Summary:
This exposes the settings for histograms as node attributes. This should
allow different nodes to set 0.99 percentiles and other custom settings.

These attributes are pulled from:
https://github.com/DataDog/dd-agent/blob/master/datadog.conf.example#L97


miketheman:
Thanks! We'll merge this in to the next minor release.

PR #280, opened 2016-03-08
support disabling the agent service

hydrant25:
We use a Packer workflow to create AMIs for our apps and then promote them through non-production environments before deploying to production. In non-production environments, it is desirable to install the Datadog Agent, but leave the service stopped and disabled via an environment attribute setting. Once in launched in production, the Agent starts and is enabled.

The cookbook already had an attribute to stop the Agent, but not to control the Agent service's `:enable`/`:disable` action. This PR adds that functionality.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -125,8 +125,8 @@ default['datadog']['use_mount'] = false
 # Change port the agent is listening to
 default['datadog']['agent_port'] = 17123
 
-# Disable the agent from starting at boot
-default['datadog']['agent_disable'] = false
+# Enable the agent to start at boot
+default['datadog']['agent_enable'] = true
 
 # Start agent or not
 default['datadog']['agent_start'] = true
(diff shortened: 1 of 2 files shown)


miketheman:
Hi @hydrant25 !

Thanks for taking the time to send this.

I think the idea is sane, but the naming and logical selection seems to be in direct opposite of the other attribute used to control the service behavior.

`default['datadog']['agent_start'] = true` with a default of true

I'd recommend something similar, like `default['datadog']['agent_enable'] = true` also with a default of true, so as to keep some consistency around the action semantics.

`agent_boot_startup` also seems very specific to your use case, whereas the `enable` action is clear.

Can you take a pass at renaming?

hydrant25:
I took a shot at renaming the attribute and variable per your comment, @miketheman. How does this look?

miketheman:
@hydrant25 that's awesome, thanks!
Now that I see them both together, it looks like the older `agent_action` is a misnomer - following this style, it would be `agent_start`.
Since this name is used only internally - the attribute already has the correct name - could you change that one too?

hydrant25:
@miketheman, ask and ye shall receive!

PR #283, opened 2016-03-14
Adding support for Marathon

cobusbernard:
Adds Marathon support.


templates/default/marathon.yaml.erb
```diff
@@ -0,0 +1,8 @@
+# Generated by Chef, local modifications will be overwritten
+
+<%= JSON.parse(({ 'instances' => @instances }).to_json).to_yaml %>
+
+init_config:
```
miketheman: How about this approach?
https://github.com/DataDog/chef-datadog/blob/10a9bfcf03e98ff2bff9f34b7321dee9cd29e2c6/templates/default/docker_daemon.yaml.erb

cobusbernard: That would be a bit better, but would argue that keeping it the same as the Mesos configuration file is more important for consistency. Unless we refactor that as well as part of this PR?

miketheman:
Hi @cobusbernard ! Would you consider porting the Kitchen test to a Spec test? See [here](https://github.com/DataDog/chef-datadog/pull/295) for an example.

Specs are much "cheaper" to run overall, and validate template renderings just as much as the kitchen tests do. My hope is that we'll port over all existing kitchen tests to specs, and some of the effort has started, so I'd like to not introduce more if possible.

AliDatadog:
Closed for inactivity

PR #284, opened 2016-03-16
add custom metrics for postgres

flah00:
* Add support for custom metrics, in postgres integration
* Add test for postgres integration


miketheman:
Hi @flah00 !
Would you consider porting the Kitchen test to a Spec test? See [here](https://github.com/DataDog/chef-datadog/pull/295) for an example.

Specs are much "cheaper" to run overall, and validate template renderings just as much as the kitchen tests do. My hope is that we'll port over all existing kitchen tests to specs, and some of the effort has started, so I'd like to not introduce more if possible.

Rebase off `master` to ensure the specs run post-commit.

flah00:
sure thing, hopefully i can get to it next week

olivielpeau:
Thanks @flah00! We'll include this in the next minor release of the cookbook.

Issue #226, opened 2015-09-03
Allow passing a custom hostname to chef-handler-datadog


olivielpeau:
Was discussed here https://github.com/DataDog/chef-handler-datadog/issues/41#issuecomment-41585416

It would be useful to find a sensible way to pass a custom hostname to the handler while preserving the priority of the ec2 flag. Opening this issue because a customer is not on ec2 and may need this feature.

miketheman:
This would likely live in https://github.com/DataDog/chef-datadog/blob/master/recipes/dd-handler.rb#L34-L38

olivielpeau:
Fixed with #308. Closing

PR #290, opened 2016-04-08
Add option to use agent hostname as handler's reported host

olivielpeau:
Tries to address #226

When the `use_agent_hostname` attribute is set to `true`, we pass the agent's hostname to the handler so that it reports with the agent hostname.

This attribute would allow users not to have duplicate hosts when the hostname of the agent is set to a custom value while the `node.name` and the ec2 instance id are not part of the host aliases sent by the agent.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -115,6 +115,10 @@ default['datadog']['check_freq'] = 15
 # More information available here: http://docs.datadoghq.com/hostnames/#agent
 default['datadog']['hostname'] = node.name
 
+# If a hostname is set for the agent, and if true,
+# use the agent hostname as the host identifier for chef-handler
+default['datadog']['use_agent_hostname'] = false
+
 # If running on ec2, if true, use the instance-id as the host identifier
 # rather than the hostname for chef-handler.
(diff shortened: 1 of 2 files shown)


miketheman:
How does this relate to #241 ?

olivielpeau:
Interesting, hadn't seen that PR. The present PR wouldn't work well when the cookbook doesn't force a value on the agent's hostname.

Basically the problem I can see with #241 and the present PR is that if we leave the hostname resolution to the agent then there's no way to pass the agent's hostname to the handler, so the option I've added wouldn't guarantee that the handler reports with the same hostname as the agent's.

Instead of the current approach, what we could do for the next major release (along with #241) would be to always force the hostname reported by the handler to `node['datadog']['hostname']` when its value is not `nil`. This would make the handler report with the `node.name` or the EC2 instance ID by default , which should match the agent's hostname in most cases, and in the few cases where the agent doesn't report with the same hostname it would allow users to set the `node['datadog']['hostname']` explicitely, which would make both reported hostnames match.

olivielpeau:
Closing in favor of #308

Issue #270, opened 2016-01-29
Update chef_handler dependency to remove pessimistic version constraint?


hydrant25:
I'm wondering if it's possible to remove the version constraint on `chef_handler` `~> 1.1.0` from this cookbook so that it pulls in the latest, v1.2.0? 

In my chef-repo, I believe this constraint, when combined with other community cookbook constraints, is causing berkshelf to resolve to using `datadog` cookbook version v0.1.2 along with `chef_handler` v1.2.0. I believe v0.1.2 was the last release the `datadog` cookbook had just a >= constraint on `chef_handler`.

Or, could we update the version constraint to `~> 1.2.0`?

darron:
Hey @miketheman - remember when that happened to us? I can't remember some of the backstory - but am sure you do.

miketheman:
@hydrant25 That's a great question.

Chef Handler 1.2.0 [broke backwards compat](https://github.com/chef-cookbooks/chef_handler/blob/master/CHANGELOG.md#v120-2015-06-25) [without a major bump](https://github.com/chef-cookbooks/chef_handler/issues/34) - this caused some hellfire to rain for a while when we tried to debug it internally.

I suspect that we'd be "ok" to remove the constraint from this cookbook, but haven't been able to truly test it, since we (and many users of this cookbook) are still on Chef 11.

I think the reasoning that you're getting datadog 0.1.2 is because that was the last version of this cookbook (oh, the early, optimistic days) where we allowed "anything greater than...", so the depsolver is getting that in conjunction with the chef_handler cookbook.

My understanding is that if you reordered your Berksfile to place `datadog` cookbook before the `chef_handler` cookbook,  the resolution would retrieve the chef_handler 1.1.9 cookbook without any issue.

hydrant25:
Thanks for the quick response, @miketheman! I did pull the constraint in my local fork and ran `kitchen test`. Looks like I ended up hitting `Timeout::Error` on starting `service[datadog-agent]` on Windows 2012r2-11186. I don't know/think that's related to the chef_handler version though.

At any rate, I did solve this in my `Berksfile` by explicitly asking for `cookbook 'datadog', version: '~> 2.2.0'`.

PR #291, opened 2016-04-11
Allow non-breaking updates of chef_handler

eherot:
Reading back to #25 when this constraint was first introduced, it doesn't seem like there's any compelling reason not to allow future non-breaking changes to come in automatically. However, refusing to allow anything above 1.1.x definitely causes problems for my other cookbooks that also need to use `chef_handler` (and there are several).

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -22,7 +22,7 @@ issues_url       'https://github.com/DataDog/chef-datadog/issues' if respond_to?
 end
 
 depends          'apt' # We recommend '>= 2.1.0'. See CHANGELOG.md for details
-depends          'chef_handler', '~> 1.1.0'
+depends          'chef_handler', '~> 1.1'
 depends          'windows'
 depends          'yum'
 


miketheman:
I had reported a breaking change that made it into 1.2.0 of chef_handler here: https://github.com/chef-cookbooks/chef_handler/issues/34

I'm not entirely convinced yet what the right version is - especially  if we want to continue to provide maximum compatibility.

Is there a particular feature from chef_handler that's needed?

eherot:
Sorry to leave this hanging for so long. There is not a particular feature that's needed. The problem is that it's very difficult to get such a specific version match to play nice with other cookbooks in the dependency tree (many of which are not maintained by me).

mfischer-zd:
What was the breaking change?

olivielpeau:
@mfischer-zd For reference see also #270 

Basically the `1.2.0` release of `chef-handler` is compatible with Chef 12+ only whereas the previous 1.x releases were compatible with Chef 10+.

We need to test that the current version of `chef-handler` (`1.4.0`), which is advertised to work with Chef 11+, works well with this cookbook before loosening the version constraint. I'll make sure to prioritize this since the version constraint seems to be blocking for many users of the cookbook.

mfischer-zd:
While I think your desire to prevent breakage for folks running on Chef < 12 is high-minded, it's not the responsibility of this cookbook to prevent such breakage.  That's an interaction between chef-handler and Chef, not you and chef-handler; and this cookbook should not intervene. 

If the combination of this cookbook, the chef-handler cookbook, and Chef 12 work fine, then I would recommend noting the issue with Chef < 12 in the README, and leaving it at that.

olivielpeau:
@mfischer-zd: Makes a lot of sense, thanks. I'll spend some time testing the recipe with chef-handler 1.2/1.3/1.4, note the issue in the README, and include that in the next minor release.

We should be able to release it by early next week.

volker-fr:
Still limited to version 1.1.x... even the current comment indicates that a higher version is recommended:

```
depends 'chef_handler', '~> 1.1' # We recommend '~> 1.3' with Chef < 12. See CHANGELOG.md for details
```
See https://github.com/DataDog/chef-datadog/blob/master/metadata.rb#L25

olivielpeau:
@volker-fr `~> 1.1` basically means `['>= 1.1', '< 2.0']` (see http://guides.rubygems.org/patterns/#declaring-dependencies)

You should try updating your cookbook dependencies (if you're using `berks`, `berks update chef_handler` should do the trick)

volker-fr:
@olivielpeau: thanks, mixed it up with ```~> 1.1.1``` which wouldn't allow versions such as 1.2

PR #293, opened 2016-04-14
Added ability to obtain API key from run_state var

hartfordfive:
I've updated the README file, although feel free to modify that to what ever you find more appropriate.


miketheman:
@hartfordfive Have you read this post on removing attributes? http://jtimberman.housepub.org/blog/2014/12/24/quicktip-delete-attributes/

If you're using an up-to-date version of Chef, you _should_ be able to use `node.rm(...` somewhere in your wrapper recipe before the end of the Chef run.

hartfordfive:
@miketheman I would have preferred to have the datadog cookbook handle the logic around this operation although the work-around you've sent me seems fairly reasonable.  I'll use that method so let's just close this PR.

PR #296, opened 2016-04-17
Allow for a hash or string when configuring tags

martinisoft:
A pattern we're employing with our wrapper cookbooks is to compose the datadog tags attribute so they're rendered out properly in the datadog template. The way it currently is now, the template is rendered at compile time which prevents wrapper cookbooks from modifying the tags during run time. By extracting the attribute to a template variable, we delay the rendering of the template variable until the run phase.

With this patch you will be able to do the following in a recipe:
```ruby
node.default['datadog']['tags'] = node['datadog']['tags'].merge('region' => 'us-east')
```
What this does is composes the current set of tags defined in the attribute with an additional tag by way of merging in a new key and value into the hash. You can still provide a string instead which is the currently shipped default so these changes are fully backward compatible.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -31,6 +31,9 @@ default['datadog']['application_key'] = nil
 default['datadog']['url'] = 'https://app.datadoghq.com'
 
 # Add tags as override attributes in your role
+# This can be a string of comma separated tags or a hash in this format:
+# default['datadog']['tags'] = { 'datacenter' => 'us-east' }
+# Thie above outputs a string: 'datacenter:us-east'
 # When using the Datadog Chef Handler, tags are set on the node with preset prefixes:
 # `env:node.chef_environment`, `role:node.node.run_list.role`, `tag:somecheftag`
 default['datadog']['tags'] = ''
--- a/templates/default/datadog.conf.erb
+++ b/templates/default/datadog.conf.erb
@@ -15,10 +15,10 @@ autorestart: <%= node['datadog']['autorestart'] %>
 skip_ssl_validation: <%= node['datadog']['web_proxy']['skip_ssl_validation'] %>
 <% end -%>
 
-<% if @tags.respond_to?(:each_pair) -%>
-tags: <%= @tags.reject{|k,v| v.empty? }.map{|k,v| "#{k}:#{v}" }.join(",") %>
+<% if node['datadog']['tags'].respond_to?(:each_pair) -%>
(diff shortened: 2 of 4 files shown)


miketheman:
Interesting. I thought the template variable was timed differently. 
Had you seen the approach suggested here? https://github.com/DataDog/chef-datadog/pull/186

This might solve the problem a little more elegantly.

martinisoft:
@miketheman Not entirely. A more flexible solution could be for adding a resource for configuring the agent that allows a wrapping cookbook to format the template however they choose.

martinisoft:
@miketheman I've added support for a hash similarly to #186 with test coverage so it shouldn't be a breaking change.

martinisoft:
I have updated the example int he initial post so this is ready for review and merge. 👍

martinisoft:
Well this has come full circle now so this more or less duplicates the work in #186 though mine can be merged and passes CI.

attributes/default.rb
```diff
@@ -31,6 +31,9 @@
 default['datadog']['url'] = 'https://app.datadoghq.com'
 
 # Add tags as override attributes in your role
+# This can be a string of comma separated tags or a hash in this format:
+# default['datadog']['tags'] = { 'datacenter' => 'us-east' }
+# Thie above outputs a string: 'datacenter:us-east'
```
miketheman: nit: typo

miketheman:
Thanks for your persistence! Can you add a spec test to cover the existing tags as a string behavior?

martinisoft:
@miketheman done! Also added a bonus spec to cover an empty tag not being output.

templates/default/datadog.conf.erb
```diff
@@ -15,7 +15,11 @@ autorestart: <%= node['datadog']['autorestart'] %>
 skip_ssl_validation: <%= node['datadog']['web_proxy']['skip_ssl_validation'] %>
 <% end -%>
 
+<% if node['datadog']['tags'].respond_to?(:each_pair) -%>
+tags: <%= node['datadog']['tags'].reject{|k,v| v.empty? }.map{|k,v| "#{k}:#{v}" }.join(',') %>
```
miketheman: Style nitpicks: spacing and unused block arg representation.
Should be `node['datadog']['tags'].reject { |_k, v| v.empty? }.map { |k, v| "#{k}:#{v}" }.join(',')`

martinisoft:
@miketheman Done :)

miketheman:
Thanks! Putting in merge queue.

miketheman:
@martinisoft Thanks for working through this! Now merged to master.

PR #298, opened 2016-04-21
Added recipe for go_expvar metrics collection

hartfordfive:


miketheman:
Hi @hartfordfive !

Can you please take a look at the approach for spec testing (vs kitchen) for integrations here: #295.

Also, we've adopted a very flexible method for templates, please follow this example: https://github.com/DataDog/chef-datadog/blob/4e59a32c7b692e8d4cf99bfb927af1de9e2c0118/templates/default/docker_daemon.yaml.erb

This leaves any further updates to the template to be attributes only, controlled by the end user, with no updates to the cookbook.

hartfordfive:
HI @miketheman,  I'll make the necessary modifications and then push the updated code.  Thanks.

hartfordfive:
@miketheman from what I can tell, the current CircleCI failure doesn't seem to be caused by anything in modifications, unless i'm missing something?

miketheman:
@hartfordfive I think it's because your branch is off the last release 2.2.0 - https://github.com/DataDog/chef-datadog/commit/4e59a32c7b692e8d4cf99bfb927af1de9e2c0118 instead of current `master` branch.

A rebase should solve it.

miketheman:
@hartfordfive Thanks! I squashed the commits and merged to master.

PR #301, opened 2016-04-29
Add support for JBoss/WildFly

ianrossi:
Since JBoss needs the `custom_jar_paths` at minimum, to work, please consider this pull request. Thank you.


miketheman:
Hi @ianrossi !

I don't know if jboss is a currently supported Datadog Agent check - can you point out where this is coming from?

Also, we cannot accept pull requests that change the cookbook's version number.

ianrossi:
Hi @miketheman, I should have put this in the initial PR, but here is a link to the DD docs page where there are instructions for JBoss/WildFly monitoring, about 3/4 down. This PR just adds the capability to put that additional config in place. I removed the version change. I think this would be helpful for anyone wanting to monitor JBoss or Wildfly with Datadog. Let me know if you have any more feedback and I'll be happy to make the changes.

ianrossi:
Replaced by [this PR](https://github.com/DataDog/chef-datadog/pull/303).

PR #302, opened 2016-05-01
Fix check for `collect_function_metrics` key

kellym:
This fixes the postgres template file when when checking for `collect_function_metrics`.


miketheman:
Hi @kellym ! Thanks for the fix.

We've been (slowly) porting over template generation slow integration tests to speedy spec tests - see #295 for an example.
Could you take a pass at converting the postgres one to support this change?

PR #278, opened 2016-02-29
Add collect_function_metrics to postgres check yml

isaacdd:

PR #304, opened 2016-05-03
Fix postgres template syntax and parameter output

### LINKED ISSUE #300: TemplateError in postgres.yaml
Seeing this error when running chef-client on a host with the postgres check enabled:
```
      Error executing action `create` on resource 'template[/etc/dd-agent/conf.d/postgres.yaml]'
      
      Chef::Mixin::Template::TemplateError
      ------------------------------------
      Undefined node attribute or method `keys?' on `node'
      
      Resource Declaration:
      ---------------------
      suppressed sensitive resource output
      
      Compiled Resource:
      ------------------
      suppressed sensitive resource output
      
      Template Context:
      -----------------
      on line #28
       26:         <% end -%>
       27:     <% end -%>
       28:     <% if i.keys?("collect_function_metrics") -%>
       29:     collect_function_metrics: i["collect_function_metrics"]
       30:     <% end -%>
```

This looks to be a result of #278 .  The `i` object here is an [ImmutableHash](http://www.rubydoc.info/github/opscode/chef/Chef/Node/ImmutableMash) so we probably want to run `i.key?('collect_function_metrics')` here.

miketheman:
- Use correct `.key?` syntax, Fix #300
- Emit correct boolean, introduced in #278
- Add postgres spec test

Closes #302

Signed-off-by: Mike Fiedler <miketheman@gmail.com>

### DIFF
--- a/spec/integrations/postgres_spec.rb
+++ b/spec/integrations/postgres_spec.rb
@@ -1,0 +1,57 @@
+describe 'datadog::postgres' do
+  expected_yaml = <<-EOF
+    instances:
+    - host: localhost
+      port: 5432
+      ssl: true
+      username: datadog
+      password: somepass
+      dbname: my_database
+      tags:
+      - spec
+      relations:
+      - apple_table
+      - orange_table
+      collect_function_metrics: true
+    init_config:
+  EOF
+
+  cached(:chef_run) do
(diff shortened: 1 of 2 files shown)


sethrosenblum:
👍 LGTM

Issue #275, opened 2016-02-12
kafka.yaml.erb - replication stats have wrong domain


gzysk8:
Metrics `kafka.replication.leader_elections` and `kafka.replication.unclean_leader_elections` are listed under the `kafka.server` domain but should be `kafka.controller` per https://github.com/DataDog/dd-agent/blob/master/conf.d/kafka.yaml.example

Using the default config as is results in:

```
Not Matching: Bean name: "kafka.controller":type="ControllerStats",name="UncleanLeaderElectionsPerSec" - Attribute name: MeanRate - Attribute type: double 
```

rlgomes:
Seems like something that got updated [here](https://github.com/DataDog/dd-agent/blob/master/conf.d/kafka.yaml.example) but the template files in the recipe weren't updated. I'm going to bring this up with their support to see if they can keep these two in sync.

olivielpeau:
Same issue as what #257 addresses, will keep this issue open until we fix it

olivielpeau:
Resolved by #340

PR #310, opened 2016-05-23
updating kafka template

rlgomes:
Noticed that the kafka template wasn't updated when the *.yaml.example* was updated in the dd-agent repo and this PR would fix https://github.com/DataDog/chef-datadog/issues/275


olivielpeau:
Hi @rlgomes! This PR is a duplicate of #257, and we'd like to rework it so that it offers a versioned template (or something similar) to allow choosing between the >=0.8.2-compatible config and the older one.

I'm going to close this PR since it's a duplicate, feel free to respond on #257, thanks!

PR #313, opened 2016-06-09
Add dogstatsd_target to template

jcftang-r7:
This is needed to tell the dogstatd component to send the data to the right place if the default agent port is changed to something else other than the 17123 value


olivielpeau:
Looks good @jcftang-r7, thanks!

Planning to merge this PR for the next minor release.

PR #317, opened 2016-06-27
[dd-agent] add multiple enpoints/api_keys conf

degemer:


spec/dd-agent_spec.rb
```diff
@@ -295,6 +295,30 @@ def set_env_var(name, value)
...
+
+    context 'does accept other_api_keys' do
+      cached(:chef_run) do
+        ChefSpec::SoloRunner.new(
+          platform: 'ubuntu',
+          version: '12.04'
+        ) do |node|
+          node.set['datadog'] = {
+            'api_key' => 'somethingnotnil',
+            'other_dd_urls' => ['app.example.com', 'app.datadoghq.com'],
```
olivielpeau: nitpick since it's not really what's tested here: it'd be better to use the full urls with the protocol, i.e. `https://app.example.com`

attributes/default.rb
```diff
@@ -181,6 +181,10 @@
 default['datadog']['histogram_aggregates'] = 'max, median, avg, count'
 default['datadog']['histogram_percentiles'] = '0.95'
 
+# Multiple endpoints/api_keys settings
```
olivielpeau: What do you think about specifying in the comment which version of the Agent these parameters will start working with? (we never do it in this file but it would be helpful IMO)

olivielpeau:
Added a few nitpicks but this is👌 

Let's merge it for `2.5.0`

recipes/dd-agent.rb
```diff
@@ -64,8 +69,8 @@
     mode 0640
   end
   variables(
-    :api_key => node['datadog']['api_key'],
-    :dd_url => node['datadog']['url']
+    :api_key => api_keys,
+    :dd_url => dd_urls
```
olivielpeau: Wouldn't it make sense to update the var names too?

recipes/dd-agent.rb
```diff
@@ -53,6 +53,11 @@
 #
 raise "Add a ['datadog']['api_key'] attribute to configure this node's Datadog Agent." if node['datadog'] && node['datadog']['api_key'].nil?
 
+api_keys = node['datadog']['api_key']
+dd_urls = node['datadog']['dd_url']
```
olivielpeau: this node attribute was previously named `node['datadog']['url']`:
- is the change intentional?
- if so are we sure that it's something that's only used by us internally? (I think so, but just to be extra sure)

olivielpeau: actually it's probably not intentional since you're using `node['datadog']['url']` in `recipes/dd-handler.rb`, so let's fix it :)

olivielpeau:
Added some comments, let me know when the tests are updated ;)

Overall it looks way better now compared to the first draft 🚀

attributes/default.rb
```diff
@@ -21,9 +21,11 @@
 # The Datadog api key to associate your agent's data with your organization.
 # Can be found here:
 # https://app.datadoghq.com/account/settings
+# This can be a list of api keys
```
olivielpeau: (just to be more explicit, I'd explain here what a list of api keys does, i.e. that it configures the Agent and the handler to send data to multiple accounts)

recipes/dd-agent.rb
degemer: Indeed. 🎆

recipes/dd-agent.rb
degemer: Typo. 😢

attributes/default.rb
degemer: You're right, it's not clear enough. I also changed list for array, as a list could be understood as a string `"apikey1, apikey2"`.

degemer:
💚  CI finally, ready for final review !

attributes/default.rb
```diff
@@ -21,9 +21,13 @@
 # The Datadog api key to associate your agent's data with your organization.
 # Can be found here:
 # https://app.datadoghq.com/account/settings
+# This can be an array of api keys (to configure the agent
+# data to send to multiple accounts)
```
olivielpeau: should be `to send data to multiple accounts`?

olivielpeau:
Added one last comment about a comment, once that's fixed let's :shipit: !

degemer:
As we discussed, I updated this PR to use a `extra_endpoints` @olivielpeau. Sorry for the multiple changes, one more review for you!

recipes/dd-handler.rb
```diff
@@ -46,11 +46,20 @@
   ENV['DATADOG_PROXY'] = proxy_url.to_s
 end
 
+extra_endpoints = []
+node['datadog']['extra_endpoints'].each do |_, endpoint|
+  next unless endpoint['enabled']
+  endpoint.delete('enabled')
```
olivielpeau: This will attempt to mutate `node['datadog']['extra_endpoints']` too, which chef doesn't allow from a recipe (it'll cause a compile error).

I think we should either copy the `endpoint` hash first (keeping in mind that it's not a regular hash but an [`ImmutableMash`](http://www.rubydoc.info/github/opscode/chef/Chef/Node/ImmutableMash) if I remember correctly) or "manually" assign the interesting fields of `endpoint` to a new hash.

Also, unless the resulting values of `extra_endpoints` var are [Mash](http://www.rubydoc.info/gems/chef/Mash)es I think their keys need to be string literals (`:url` instead of `'url'`) since that's how we access the values in the handler.

attributes/default.rb
```diff
@@ -26,6 +26,13 @@
 # Create an application key on the Account Settings page
 default['datadog']['application_key'] = nil
 
+# Use this attribute to send data to multiple accounts
```
olivielpeau: nitpick: `multiple` -> `additional`? Just want to make extra-clear that these attributes are used to configure accounts on top of the "default" one.

Also, do you think it's worth mentioning that this affects the Agent _and_ the datadog handler>

olivielpeau:
Had 2 comments on the PR

recipes/dd-handler.rb
olivielpeau: about my comment on the `Mash` thing, not sure it's super important since in the handler we copy the config to a `Mash` since v0.5.0 (see https://github.com/DataDog/chef-handler-datadog/pull/50)

I'd be surprised if anyone still uses `chef-handler-datadog` < `0.5.0`...

degemer:
Comments addressed, and a few tests added.
It's a start, because I didn't want to spend too much time on it. We should add more "one day".

PR #319, opened 2016-06-30
updated kafka config to match documentation

joshcody:
Update kafka collector template to match the documentation.


olivielpeau:
Thanks @joshcody for your PR, unfortunately this is a duplicate of #257, so I'm going to close your PR. We're going to try to allow using either the old config or the new one depending on the version of kafka that needs to be monitored.

This request is getting quite some traction so I'll make sure to prioritize it. Thanks!

Issue #306, opened 2016-05-06
IIS integration is not collecting data


clmoreno:
I installed datadog and added the iis integration using the recipe datadog::iis, but it is not collecting data..

`node.default.datadog.iis.instances= [{"host" => "localhost", "tags" => ["tag","tag2"]}]`
`include_recipe "datadog::iis"`

also tried

`node.default.datadog.iis.instances= [{"host" => "localhost", "tags" => ["tag","tag2"]}]`
`datadog_monitor 'iis' do`
`instances node['datadog']['iis']['instances']`
`end`

![datadog_iiis](https://cloud.githubusercontent.com/assets/12781407/15074585/fe151106-135c-11e6-9a57-69d9e19bce70.png)

what is the correct way to add the integration?

clmoreno:
This issue was fixed with the merge #324

PR #324, opened 2016-07-11
Support for iis wmi provider

clmoreno:
The new Datadog agent has been released and with it comes the solution for the issue #306 

This pull request include the support to use this new feature adding "provider" to the IIS attributes declaration, as follows.

`node.datadog.iis.instances = [
	{
		"host" => "localhost",
		"provider" => '64',
		"tags" => ["tag1", "tag2"],
		"sites" => ["Default Web Site"]
	}
]
`

### DIFF
--- a/templates/default/iis.yaml.erb
+++ b/templates/default/iis.yaml.erb
@@ -19,6 +19,7 @@ instances:
     username: <%= i['username'] %>
     password: <%= i['password'] %>
     <% end -%>
+    provider: <%= i['provider'] %>
     <% if i.key?('tags') -%>
     tags:
       <% i['tags'].each do |t| -%>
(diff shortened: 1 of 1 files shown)


olivielpeau:
Hi @clmoreno and thanks for your contribution!

The tests are failing (see https://travis-ci.org/DataDog/chef-datadog/builds/144015476) because the `provider` is set in the `yaml` file even when the attribute is not defined.

Could you add an `<% if i.key?('provider') -%>` in the template so that the parameter is set in the yaml only when the attribute exists? Thanks!

clmoreno:
@olivielpeau fixed!

olivielpeau:
Looks good thanks! We'll include this in the next minor release

Issue #311, opened 2016-06-03
yum gpg key should be an attribute


icberg7:
Our servers are behind a firewall, and we use a proxy/mirror to make the RPM packages available internally. We also need to host the GPG keys internally somewhere as well, so we really need to be able to configure the gpg key through a wrapper or something.

olivielpeau:
Hi @icberg7, such a feature would be very welcome, would you mind opening a PR implementing it? (as you seem to have implemented it on your fork of the cookbook)

Thanks!

iancward:
Hi @olivielpeau I had to get approval to submit, but now that it's approved, please see #326.

Thanks.

PR #326, opened 2016-07-13
Make yum repo GPG key an attribute

iancward:
This fixes GH-311.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -62,10 +62,14 @@ architecture_map = {
 }
 architecture_map.default = 'x86_64'
 
+# Older versions of yum embed M2Crypto with SSL that doesn't support TLS1.2
+prefix = node['platform_version'].to_i < 6 ? 'http' : 'https'
+
 default['datadog']['installrepo'] = true
 default['datadog']['aptrepo'] = 'http://apt.datadoghq.com'
 default['datadog']['aptrepo_dist'] = 'stable'
 default['datadog']['yumrepo'] = "http://yum.datadoghq.com/rpm/#{architecture_map[node['kernel']['machine']]}/"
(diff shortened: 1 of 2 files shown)


olivielpeau:
Thanks @iancward, this looks good! We'll include it in the next minor release.

attributes/default.rb
```diff
@@ -62,10 +62,14 @@
 }
 architecture_map.default = 'x86_64'
 
+# Older versions of yum embed M2Crypto with SSL that doesn't support TLS1.2
+prefix = node['platform_version'].to_i < 6 ? 'http' : 'https'
```
olivielpeau: After giving it some thought, I think this deserves a more accurate condition since it will run on all platforms. Could you change this to:

```ruby
yum_protocol = 'https'
if node['platform_family'] == 'rhel' && node['platform_version'].to_i < 6
  # Older versions of yum embed M2Crypto with SSL that doesn't support TLS1.2
  yum_protocol = 'http'
end
```

(I've also renamed `prefix` `yum_protocol` since that name makes more sense here IMO)

olivielpeau:
@iancward Just added a comment actually, let me know what you think, thanks!

attributes/default.rb
iancward: I merely migrated the code from the repository recipe, but I just noticed that it's inside of a `case node['platform_family']` block. 

This is fixed in6eb91d7. I did it a bit differently to try to be more DRY.

olivielpeau:
Thanks @iancward, looks better now, I think this is ready for the next minor release

irabinovitch:
@iancward Thanks for your efforts on this PR. We're currently working through a QA/testing cycle on the next release, which we expect to include this PR.

PR #329, opened 2016-07-15
Defer evaluation of api key attribute

### LINKED ISSUE #328: Use lazy evaluation for node attributes rendered into templates
Assume a user wants to store the API key in an external information service other than Chef server, such as Amazon S3 or Hashicorp Vault.  This key would be fetched during the run phase, not the compile phase; and therefore the attribute cannot be set during the compile phase as with other attributes.

So, use Chef's `lazy` (delayed evaluation) operator where possible in any resource attributes that render files.

mfischer-zd:
Some users may wish to assign the `api_key` attribute during runtime
instead of compile time -- particularly, users who want to obtain the
value of the key from an external service. Allow doing so by deferring
evaluation of the `api_key` attribute.

Closes #328

### DIFF
--- a/attributes/cassandra.rb
+++ b/attributes/cassandra.rb
@@ -1,1 +1,0 @@
-default['datadog']['cassandra']['version'] = 1
--- a/recipes/cassandra.rb
+++ b/recipes/cassandra.rb
@@ -35,6 +35,8 @@ include_recipe 'datadog::dd-agent'
 # }
 # ```
 
+node.default['datadog']['cassandra']['version'] = 1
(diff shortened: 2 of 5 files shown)


olivielpeau:
Thanks @mfischer-zd for this PR and the explanation you provided, we'll make sure to review this soon.

mfischer-zd:
Can I get a status update, please?

mfischer-zd:
Ping!

mfischer-zd:
Closing in favor of #345

PR #330, opened 2016-07-19
Use HTTPS for yumrepo when applicable

aknarts:


aknarts:
Would be even easier change if done in https://github.com/DataDog/chef-datadog/pull/326

olivielpeau:
Thanks for the PR @aknarts!

This PR changes the value assigned to an attribute by the cookbook, and even though the change is minor I think it still counts as something that should be released as part of a major version (see the [Cookbook versioning policy](https://chef-community.github.io/cvp/),  )

I'm going to slate this for the next major version, let me know what you think.

As an aside, even without this change I don't think there should be any security concern around downloading the packages through plain http since the packages are signed and the public key is pulled through https (when the platform allows it).

aknarts:
Hi, there should be no reason to not use HTTPS if at all possible. And it is now even easier since this got merged https://github.com/DataDog/chef-datadog/pull/326/files#diff-25e5d4a4446ae12a0d6f1162b6160375R66 
Line 76 would just need to change the same way as the line 77 did. i.e. 
`default['datadog']['yumrepo'] = ""#{yum_protocol}://yum.datadoghq.com/rpm/#{architecture_map[node['kernel']['machine']]}/"`

As a result this merge is redundant and I am closing it(will create a new one with the proposed change.

PR #332, opened 2016-07-22
Enhance remove-dd-agent.rb 

raycrawford:
Patch to fix or enhance remove-dd-agent.rb's functionality to include Windows. Tested on Windows 2012 R2 and Windows 2008 R2 SP1. In response to open issue 331 (DataDog#331).

Prior Linux-only code was included in case logic unchanged.

### DIFF
--- a/recipes/remove-dd-agent.rb
+++ b/recipes/remove-dd-agent.rb
@@ -1,4 +1,11 @@
 # Run this recipe to completely remove the Datadog Agent
-package 'datadog-agent' do
-  action :purge
+case node['os']
+when 'linux'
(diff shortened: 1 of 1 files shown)


olivielpeau:
Looks good, thanks @raycrawford!

olivielpeau:
NB: on Windows, this appears to work only on Chef >= 12.6 (it uses the `windows_package` resource defined in core Chef, not the one in the `windows` cookbook)

I've added a note in 962821863c , we may want to investigate if we could extend it to Chef 12.x < 12.6 in the future.

PR #333, opened 2016-07-27
direct recipe support for sqlserver configuration

smpilo:
This update would allow you to use recipe['datadog::sqlserver'] and configure both instances and custom_metrics.

### DIFF
--- a/providers/monitor.rb
+++ b/providers/monitor.rb
@@ -21,7 +21,8 @@ action :add do
     variables(
       :init_config => new_resource.init_config,
       :instances => new_resource.instances,
-      :version => new_resource.version
+      :version => new_resource.version,
+      :custom_metrics => new_resource.custom_metrics
     )
     cookbook new_resource.cookbook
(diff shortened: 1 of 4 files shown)


olivielpeau:
Hi @smpilo and thanks for opening this PR!

I don't think using a new attribute (`custom_metrics`) on the provider is the right approach here since `custom_metrics` is specific to this check and is part of the `init_config` in the check configuration.

Could you use the existing `init_config` attribute instead? (which means that the resource would be passed `node['datadog']['sqlserver']['init_config']` and that the template would test the existence of the `custom_metrics` key in `init_config` to determine whether to add them or not).

Let me know what you think, and thanks again for contributing!

smpilo:
Sure, I can see that.

I’ll take a look, thanks!

Shane

mlcooper:
@smpilo @olivielpeau I think this one should be able to be closed since #425 was merged

olivielpeau:
That's right @mlcooper, thanks, closing this one

smpilo:
@mlcooper Sweet, thanks 👍

PR #335, opened 2016-08-05
[integrations] Update integration test

olivielpeau:
So that we test that the package is installed instead of testing that
a file is present in a hardcoded folder: what's interesting to test
from the cookbook's perspective is not the exact file installed by the
package, but that the package itself is installed.

Also, remove windows platforms from `datadog_integrations` suite
(since the recipe is not meant to run on windows).


degemer:
Makes sense 👍

PR #344, opened 2016-08-12
Don't enable cassandra monitoring everywhere

mfischer-zd:
Attribute to enable cassandra monitoring was set on all hosts.
Only enable it when zendesk_datadog::cassandra is run.

### DIFF
--- a/attributes/cassandra.rb
+++ b/attributes/cassandra.rb
@@ -1,1 +1,0 @@
-default['datadog']['cassandra']['version'] = 1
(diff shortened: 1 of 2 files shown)


olivielpeau:
@mfischer-zd: Just to clarify, the current behavior of the cookbook is to always set a default to `node['datadog']['cassandra']['version']` but that alone doesn't enable the cassandra integration (the `datadog::cassandra` recipe has to be run to actually enable it).

Could you describe what's the exact issue you've run into?

mfischer-zd:
It was an interaction between the value being set by default (regardless of whether the `datadog::cassandra` recipe was run) and a switch in one of our wrapper cookbooks that did an `include_recipe` of the recipe if the attribute was set.  

I've fixed our cookbook logic so I'm going to close this issue, since I don't think it's worth changing the Datadog cookbook over.

Issue #341, opened 2016-08-11
Provide alternative methods of setting application and API keys


mfischer-zd:
Using node attributes to set the application and API keys has security implications that may be unsuitable for more security-conscious users.  For example, such attributes will automatically be saved back to the Chef server unless you blacklist them somehow (e.g., via the `whitelist` parameters in `client.rb`, or by using a third-party cookbook like blacklist-node-attrs that monkey-patches `node.save`).  It also encourages putting such keys in source control, which should be discouraged for sensitive information such as this.

I'd recommend that in addition to the use of `lazy` (see #328), it be possible to fetch API and application keys via the `node.run_state` Hash that are put there by recipes that execute prior to the Datadog recipes.  If not present, the code can fall back to using the attributes.

Issue #338, opened 2016-08-08
Unable to test my cookbook w/ ChefSpec when including datadog::dd-handler


jeffbyrnes:
I’m using `datadog` v2.5.0.

I have a cookbook, with a recipe that:

``` ruby
include_recipe 'datadog::dd-agent'
include_recipe 'datadog::dd-handler'
```

However, this is the output I get:

```
6) ds_base::monitoring When all attributes are default, on Ubuntu 16.04 converges successfully
   Failure/Error: expect { chef_run }.to_not raise_error

     expected no Exception, got #<LoadError: cannot load such file -- chef/handler/datadog> with backtrace:
       # /var/folders/30/nsmp68z977s8q1x10z4h9_fr0000gn/T/chefspec20160808-17741-60xwg5file_cache_path/cookbooks/datadog/recipes/dd-handler.rb:38:in `from_file'
       # /var/folders/30/nsmp68z977s8q1x10z4h9_fr0000gn/T/chefspec20160808-17741-60xwg5file_cache_path/cookbooks/ds_base/recipes/monitoring.rb:12:in `from_file'
       # ./spec/unit/recipes/monitoring_spec.rb:24:in `block (3 levels) in <top (required)>'
       # ./spec/unit/recipes/monitoring_spec.rb:28:in `block (4 levels) in <top (required)>'
       # ./spec/unit/recipes/monitoring_spec.rb:28:in `block (3 levels) in <top (required)>'
   # ./spec/unit/recipes/monitoring_spec.rb:28:in `block (3 levels) in <top (required)>'
```

If I `chef gem install chef-handler-datadog`, things work properly, but that feels a bit janky.

olivielpeau:
Hi @jeffbyrnes and thanks for your feedback!

The `dd-handler` recipe has to `require chef/handler/datadog` because some versions of the [`chef_handler`](https://github.com/chef-cookbooks/chef_handler) cookbook need this `require` statement to work properly.

Obviously using the `require` doesn't work when the `chef-handler-datadog` gem hasn't actually been installed by the `chef_gem` resource that precedes the `require`. So `dd-handler` doesn't work for chefspec runs and for chef runs that use the `why-run` mode (see related issue #118).

[...]

PR #345, opened 2016-08-12
Update how agent and handler are configured

### LINKED ISSUE #328: Use lazy evaluation for node attributes rendered into templates

mfischer-zd:
See commits for information.  Closes #328, #341

### DIFF
--- a/recipes/dd-handler.rb
+++ b/recipes/dd-handler.rb
@@ -32,10 +32,7 @@ ENV['DATADOG_HOST'] = node['datadog']['url']
 chef_gem 'chef-handler-datadog' do # ~FC009
   action :install
   version node['datadog']['chef_handler_version']
-  # Chef 12 introduced `compile_time` - remove when Chef 11 is EOL.
-  compile_time true if respond_to?(:compile_time)
 end
-require 'chef/handler/datadog'
 
 # add web proxy from config support
 web_proxy = node['datadog']['web_proxy']
@@ -60,6 +57,8 @@ chef_handler 'Chef::Handler::Datadog' do
     end
     config
   end
+
+  require 'chef/handler/datadog'
   source 'chef/handler/datadog'
   arguments(
     if respond_to?(:lazy)
@@ -69,5 +68,11 @@ chef_handler 'Chef::Handler::Datadog' do
     end
   )
   supports :report => true, :exception => true
-  action :nothing
-end.run_action(:enable) if node['datadog']['chef_handler_enable']
+  action(
+    if node['datadog']['chef_handler_enable']
+      :enable
+    else
+      :nothing
+    end
+  )
+end
--- a/recipes/dd-handler.rb
+++ b/recipes/dd-handler.rb
@@ -46,21 +46,28 @@ unless web_proxy['host'].nil?
   ENV['DATADOG_PROXY'] = proxy_url.to_s
 end
 
-handler_config = {
-  :api_key => node['datadog']['api_key'],
-  :application_key => node['datadog']['application_key'],
-  :use_ec2_instance_id => node['datadog']['use_ec2_instance_id'],
-  :tag_prefix => node['datadog']['tag_prefix']
-}
-
-unless node['datadog']['use_ec2_instance_id']
-  handler_config[:hostname] = node['datadog']['hostname']
-end
-
 # Create the handler to run at the end of the Chef execution
 chef_handler 'Chef::Handler::Datadog' do
+  def handler_config # rubocop:disable Metrics/AbcSize
+    config = {
+      :api_key => node['datadog']['api_key'],
+      :application_key => node['datadog']['application_key'],
+      :use_ec2_instance_id => node['datadog']['use_ec2_instance_id'],
+      :tag_prefix => node['datadog']['tag_prefix']
+    }
+    unless node['datadog']['use_ec2_instance_id']
+      config[:hostname] = node['datadog']['hostname']
+    end
+    config
+  end
   source 'chef/handler/datadog'
-  arguments [handler_config]
+  arguments(
+    if respond_to?(:lazy)
+      lazy { [handler_config] }
+    else
+      [handler_config]
+    end
+  )
   supports :report => true, :exception => true
   action :nothing
 end.run_action(:enable) if node['datadog']['chef_handler_enable']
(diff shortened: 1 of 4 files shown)


olivielpeau:
Thanks @mfischer-zd, we'll spend some time in the next few weeks to test these changes in details. Being able to prevent storing the unencrypted app/api keys in Chef Server is definitely a very welcome feature.

Can you confirm that your changes on `chef_gem`'s `compile_time` attribute work well with Chef 11?

mfischer-zd:
> Can you confirm that your changes on chef_gem's compile_time attribute work well with Chef 11?

I cannot confirm; we don't use Chef 11.  (It should be noted that Chef 12 has been out for 18 months.)

mfischer-zd:
Hello, have you had time to review this?  It's been nearly 2 months.

iancward:
I'd recommend leaving the `compile_time true` in, as even though Chef 12 has been out for some time, I'm sure Chef 11 is still in use in various places, and it's still available for download at https://downloads.chef.io.

recipes/dd-handler.rb
```diff
@@ -32,10 +32,7 @@
 chef_gem 'chef-handler-datadog' do # ~FC009
   action :install
   version node['datadog']['chef_handler_version']
-  # Chef 12 introduced `compile_time` - remove when Chef 11 is EOL.
-  compile_time true if respond_to?(:compile_time)
```
haidangwa: As the comment indicated, this `compile_time` should not be removed yet. Chef 11 is still supported by Chef and has not reached EOL.

mfischer-zd:
@iancward @haidangwa It wasn't removed to stop supporting Chef 11.  It was removed because it's no longer needed with this PR.

haidangwa:
@mfischer-zd If you have tests with chef 11 clients that show it doesn't have any adverse affect, then I'd be ok with that, but removing that will change when the gem is installed. If the difference between getting installed at compile time versus getting installed at convergence doesn't have any negligible affect, then I'm ok.  If anything, using `compile_time false` would suppress warnings in clients pre-chef 12.1.

mfischer-zd:
@haidangwa I'd expect Datadog to run the tests under whatever versions of Chef it plans to support.  If the test suite still passes, then I'd expect it to be OK.

olivielpeau:
Quick update on the review of this PR: we plan to test and release these changes for the next minor release of the cookbook (2.8), in the next month or so. Sorry for the delay in getting this merged and thanks again for this big contribution.

Since the cookbook still needs to support Chef 11 the change on `compile_time` can't be released as-is and will not be merged with the rest of the changes.

Unfortunately we don't have automated tests on Chef 11, we need to improve that.

mfischer-zd:
> Since the cookbook still needs to support Chef 11 the change on compile_time can't be released as-is and will not be merged with the rest of the changes.

This part of the change should not impact Chef 11 compatibility.

iancward:
But does the gem need to be installed at compile time to work? If not, then the line can safely be removed. It it does need to be installed at compile time, then the comment should be removed but the code line left. Since Since Chef changed the install time of chef_gem from compile time to run time in Chef 12.

recipes/dd-handler.rb
```diff
@@ -46,21 +43,36 @@
...
+      :use_ec2_instance_id => node['datadog']['use_ec2_instance_id'],
+      :tag_prefix => node['datadog']['tag_prefix']
+    }
+    unless node['datadog']['use_ec2_instance_id']
+      config[:hostname] = node['datadog']['hostname']
+    end
+    config
+  end
+
+  require 'chef/handler/datadog'
```
mfischer-zd: @iancward Note the move of the `require` line to here.  Since it's now declared in a resource provider block, it isn't evaluated until runtime.  Therefore, the Gem no longer needs to be installed at compile time.  As long as it's declared before this handler resource is declared, things will work properly even if the Gem is installed at runtime.

olivielpeau:
Thanks @mfischer-zd for clarifying your changes on `compile_time`, makes sense to me now. Haven't tested it thoroughly but this change looks good.

alq666:
@olivielpeau we need an update on this.

recipes/dd-handler.rb
```diff
@@ -32,10 +32,7 @@
 chef_gem 'chef-handler-datadog' do # ~FC009
   action :install
   version node['datadog']['chef_handler_version']
-  # Chef 12 introduced `compile_time` - remove when Chef 11 is EOL.
-  compile_time true if respond_to?(:compile_time)
 end
-require 'chef/handler/datadog'
```
olivielpeau: Took some time to test this, and unfortunately moving the `require` doesn't change its evaluation time, it's still evaluated at compile time.

This commit still works in Chef 12 because not specifying `compile_time` on `chef_gem` still makes the install of `chef-handler-datadog` happen at compile time (see, in the link you provided: `The default behavior has not changed, but every `chef_gem` resource will now emit out a warning`). The default will actually change to `compile_time false` in Chef 13.

So we should keep the `compile_time true` line until we can get rid of the `require` altogether (and I _think_ we'll be able to do that once we bump the requirement on `chef_handler` to `~> 1.3`, see related https://github.com/DataDog/chef-datadog/issues/338#issuecomment-238295004 ). Let's keep this discussion in a separate issue/PR though.

mfischer-zd: OK with me.

olivielpeau:
Just opened:
- #395, a rebased version of this PR on top of the current master branch
- #396 to remove the `require` statement in `dd-handler`

alq666:
@olivielpeau thanks, keep me posted.

olivielpeau:
Going to close this PR as I've just merged #395. Will be released with `2.8.0` in the next couple of days.

Thanks again @mfischer-zd!

olivielpeau:
Update: `2.8.0` is out https://github.com/DataDog/chef-datadog/releases/tag/v2.8.0 :)

PR #347, opened 2016-08-15
datadog/chef-handler-datadog#85 - adding the ability to blacklist the…

ABrehm264:
… tags that the host sends to datadog

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -59,12 +59,11 @@ default['datadog']['create_dd_check_tags'] = nil
 # Collect EC2 tags, set to 'yes' to collect
 default['datadog']['collect_ec2_tags'] = nil
 
-# A regex used to blacklist the tags that the host sends to datadog.
+# Set this regex to exclude some Chef node tags from the host tags that the datadog handler sends to Datadog
 # https://github.com/DataDog/chef-handler-datadog/issues/85
-# This affects not only event tags but also the host-level tags.
 # This means that all the metrics and service checks coming from the
 #  host/Agent would also stop being tagged with these excluded tags.
-# EX: 'APP_.*' allows all tags except those which look like APP_.*
+# EX: 'app_.*' allows all tags except those which look like app_.*
 default['datadog']['tags_blacklist_regex'] = nil
 
 # Autorestart agent
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -59,6 +59,14 @@ default['datadog']['create_dd_check_tags'] = nil
 # Collect EC2 tags, set to 'yes' to collect
 default['datadog']['collect_ec2_tags'] = nil
 
+# A regex used to blacklist the tags that the host sends to datadog.
+# https://github.com/DataDog/chef-handler-datadog/issues/85
+# This affects not only event tags but also the host-level tags.
+# This means that all the metrics and service checks coming from the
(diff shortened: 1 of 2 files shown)


olivielpeau:
Thanks @ABrehm264!

You also need to pass the attribute to the handler in [this hash](https://github.com/DataDog/chef-datadog/blob/v2.5.0/recipes/dd-handler.rb#L49-L54).

attributes/default.rb
```diff
@@ -38,6 +38,14 @@
 # Collect EC2 tags, set to 'yes' to collect
 default['datadog']['collect_ec2_tags'] = nil
 
+# A regex used to blacklist the tags that the host sends to datadog.
+# https://github.com/DataDog/chef-handler-datadog/issues/85
+# This affects not only event tags but also the host-level tags.
+# This means that all the metrics and service checks coming from the
+#  host/Agent would also stop being tagged with these excluded tags.
+# EX: 'APP_.*' allows all tags except those which look like APP_.*
```
olivielpeau: nitpick: just for consistency with the lower case of the tags in Datadog it might be better to use a lower-cased pattern in the example. What do you think?

attributes/default.rb
```diff
@@ -38,6 +38,14 @@
 # Collect EC2 tags, set to 'yes' to collect
 default['datadog']['collect_ec2_tags'] = nil
 
+# A regex used to blacklist the tags that the host sends to datadog.
```
olivielpeau: maybe we could reword this to: `Set this regex to exclude some Chef node tags from the host tags that the datadog handler sends to Datadog`

We could then remove the next line (`This affects not only event tags but also the host-level tags.`)

Let me know if that sounds better to you or not

ABrehm264:
@olivielpeau - oh man, I'm sorry I thought I had gotten around to updating my comments to fall more in line with what you had in mind. Please take a gander at it now, it should satisfy the lower case request on the test case along with changing the description on what it does.

olivielpeau:
@ABrehm264 sorry to respond so late, but I realize now that you've missed one of my comments: the cookbook has to pass the attribute to the handler in [this hash](https://github.com/DataDog/chef-datadog/blob/v2.6.0/recipes/dd-handler.rb#L57-L64), otherwise the handler won't be configured with it ;)

Once you've changed this the PR looks good to me!

ABrehm264:
@olivielpeau, check it out now please. I think I got it. *crosses fingers*

recipes/dd-handler.rb
```diff
@@ -60,7 +60,8 @@
   :use_ec2_instance_id => node['datadog']['use_ec2_instance_id'],
   :tag_prefix => node['datadog']['tag_prefix'],
   :url => node['datadog']['url'],
-  :extra_endpoints => extra_endpoints
+  :extra_endpoints => extra_endpoints,
+  :tags_blacklist_regex => ['datadog']['tags_blacklist_regex']
```
olivielpeau: `['datadog']['tags_blacklist_regex']` should read `node['datadog']['tags_blacklist_regex']` instead

ABrehm264:
@olivielpeau - alright, that should now be done

degemer:
Hello @ABrehm264, I rebased your changes in https://github.com/DataDog/chef-datadog/pull/397, it'll go out with the next release. Thanks again for the contribution!

PR #348, opened 2016-08-15
Allow full disable of dogstatsd

ccannell67:
If default['datadog']['dogstatsd'] = false|nil, not only don't set dogstatsd attributes, but also set
use_dogstatsd: no
to fully disable dogstatsd, don't start the statsd process.
See: https://github.com/DataDog/dd-agent/blob/master/datadog.conf.example#L149


olivielpeau:
Thanks @ccannell67, this makes a lot of sense!

Looks ready for our next minor release.

PR #351, opened 2016-08-29
Use HTTPS for yumrepo when applicable

aknarts:
Updated version of https://github.com/DataDog/chef-datadog/pull/330 using the newer way of deciding which protocol to use


olivielpeau:
Thanks for the updated PR @aknarts 

After giving some thought I don't see how it could break BC, so let's merge this for the next minor release even though it slightly changes the default value of an attribute.

PR #353, opened 2016-09-07
If a cookbook is not specified, fall back to the integration template instead of failing

eherot:
This seems like it should make the resource behavior a little more intuitive and avoid having to make the resource call overly verbose.

Also I've only really tried this with my own very particular setup so please let me know if there are any use cases I've overlooked here.

### DIFF
--- a/providers/monitor.rb
+++ b/providers/monitor.rb
@@ -19,7 +19,10 @@ action :add do
       mode 00600
     end
 
-    source 'integration.yaml.erb' if new_resource.use_integration_template
+    if new_resource.use_integration_template ||
+       new_resource.cookbook == 'datadog'
+      source 'integration.yaml.erb'
+    end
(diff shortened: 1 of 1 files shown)


olivielpeau:
Hi @eherot!

Unfortunately a lot of recipes in this cookbook use specific templates (for instance, `cassandra.rb`, `kafka.rb`, etc). This change would make them use the `integration.yaml.erb` instead and the resulting `yaml`s would be invalid/incomplete.

In a future major release of the cookbook we'll try to stop using integration-specific templates and use the same basic template for all the integrations that the cookbook can set up, but until then I think using `use_integration_template` on the resource works reasonably well for your use case.

How does that sound?

eherot:
Ah ok that makes sense. Thanks for the feedback.

PR #356, opened 2016-09-13
Apply rubocop

rnt:
The cookbook do not pass our rubocop check.

This fix the problem. :)

### DIFF
--- a/providers/monitor.rb
+++ b/providers/monitor.rb
@@ -16,7 +16,7 @@ action :add do
       inherits false
     else
       owner 'dd-agent'
-      mode 00600
+      mode 0o0600
     end
 
     source 'integration.yaml.erb' if new_resource.use_integration_template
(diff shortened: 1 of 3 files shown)


iancward:
I think it may actually be better to express mode as a string: `'0600'` rather than an octal directly.

olivielpeau:
Hi @rnt!

I think we should just comply with foodcritic and the version/rules of rubocop that is commonly used in chef cookbooks (i.e. what's defined https://github.com/chef/cookstyle, it's not what's currently used on this cookbook right now but it's still a good reference).

The chef cookstyle doesn't provide guidelines about file permission octals (yet), but foodcritic does: http://www.foodcritic.io/#FC006 (and it doesn't complain about our current notation fwiw)

Maybe using a string (`'600'`) would make the most sense, as I don't find the `0o600` notation very easy to read... Thoughts?

miketheman:
My vote is for `'600'` over the other notation.

This repo also [declares its own rubocop rules](https://github.com/DataDog/chef-datadog/blob/32bef19902ce1b547321d701bc590ca34127ef5c/.rubocop.yml#L25-L27) - are they not being taken into account when you run the linter against your code?

PR #357, opened 2016-09-14
[dd-agent] Fix version test when version contains an epoch

olivielpeau:
Previously the test wouldn't give the expected result for versions like
`1:5.8.0:1` for instance (which is standard for our debian pkg).


degemer:
Nice, and still works with the old version numbers !
:shipit:

PR #358, opened 2016-09-14
[dd-agent][yum] Fix `datadog-agent-base` guard logic

olivielpeau:
We want to remove the `datadog-agent-base` pkg only if it's installed,
which is what the `rpm -q datadog-agent-base` cmd checks (it returns
`0` when the pkg is installed, `1` otherwise).


degemer:
screen shot 2016-09-14 at 11 45 28

👍

olivielpeau:
thanks @degemer! Merging

PR #359, opened 2016-09-14
[dd-agent] Add attribute to allow agent downgrade

olivielpeau:
Still not allowed by default. Switching the attribute to `true` allows downgrades:

* Straightforward for rhel/fedora: the yum and rpm package providers
support a parameter that does just that
* Uses `--force-yes` on debian because:

 1. there's no direct option on the apt
provider
 2. `apt-get` supports a more fine-grained option
(`--allow-downgrades`) since version `1.1` of `apt`, but it ships with
ubuntu starting from `16.04` only, so we stick to using `--force-yes`. It could be nice to switch the option depending on the version of `apt`, but it'd add some complexity to the logic.

### Notes on downgrading packages with `apt`

With `apt`, downgrading can be dangerous for packages that have dependencies/on which other packages depend. From `dpkg`'s manpage, on downgrades:

> Warning: At present dpkg does not do any dependency checking on downgrades  and  therefore  will not warn  you  if  the  downgrade breaks the dependency of some other package. This can have serious side effects, downgrading essential system components can even make your whole system unusable.

This shouldn't be the case of the `datadog-agent` package so downgrading should be relatively safe.

Also, using `--force-yes` is equivalent to allowing `remove-essential`, `downgrades` and `change-held-packages`. None of these should be dangerous when applied to the `datadog-agent` package only. For reference see the manpage of `apt` `1.2`: http://manpages.ubuntu.com/manpages/xenial/en/man8/apt-get.8.html

### DIFF
--- a/recipes/_install-linux.rb
+++ b/recipes/_install-linux.rb
@@ -42,5 +42,7 @@ else
   package 'datadog-agent' do
     version dd_agent_version
     action node['datadog']['agent_package_action'] # default is :install
+    allow_downgrade node['datadog']['agent_allow_downgrade'] if %w(rhel fedora).include?(node['platform_family'])
+    options '--force-yes' if node['datadog']['agent_allow_downgrade'] && node['platform_family'] == 'debian'
   end
 end
--- a/recipes/_install-linux.rb
+++ b/recipes/_install-linux.rb
@@ -39,10 +39,18 @@ else
     not_if 'apt-cache policy datadog-agent-base | grep "Installed: (none)"' if node['platform_family'] == 'debian'
   end
   # Install the regular package
-  package 'datadog-agent' do
-    version dd_agent_version
-    action node['datadog']['agent_package_action'] # default is :install
-    allow_downgrade node['datadog']['agent_allow_downgrade'] if %w(rhel fedora).include?(node['platform_family'])
-    options '--force-yes' if node['datadog']['agent_allow_downgrade'] && node['platform_family'] == 'debian'
+  case node['platform_family']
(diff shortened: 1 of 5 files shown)


degemer:
Despite the Foodcritic warning, it works. If you don't find a way to fix it, I'm fine with silencing this error in this specific case.

Tested on Ubuntu 15.04 and Centos 7.0. 👍

olivielpeau:
@degemer The Foodcritic warning kind of makes sense, so I separated the `package` providers explicitly, let me know if you think it's worth it.

Also added spec tests on the new attribute

recipes/_install-linux.rb
```diff
@@ -39,8 +39,18 @@
...
+      version dd_agent_version
+      action node['datadog']['agent_package_action'] # default is :install
+      options '--force-yes' if node['datadog']['agent_allow_downgrade']
+    end
+  when 'rhel', 'fedora'
+    yum_package 'datadog-agent' do
+      version dd_agent_version
+      action node['datadog']['agent_package_action'] # default is :install
+      allow_downgrade node['datadog']['agent_allow_downgrade']
+    end
```
olivielpeau: Discussed this offline: since we're doing the same test in [`recipes/repository.rb`](https://github.com/DataDog/chef-datadog/blob/v2.5.0/recipes/repository.rb#L20-L21), the cookbook doesn't work on other platforms anyway, so no need to fall back on a default here.

olivielpeau:
CI is 💚 , merging, thanks @degemer!

PR #365, opened 2016-09-22
Prepare pkg keys rotation

olivielpeau:
~~_WIP, PR open for discussion. Tests are failing for now, I'll fix them and add new ones once we agree that this is the right approach_~~

The idea is to make the pkg managers trust the new keys in addition to the current keys, in the next version of this cookbook (before the new keys are used). In the future, once we switch the APT repository and the RPM packages over to the new keys, we could release yet another version of the cookbook that only trusts the newer keys.

This approach ensures that, when we switch over to the new keys:
- for people using a version of the cookbook that includes this PR, the switch will be unnoticeable
- for people using an older version of the cookbook, apt pkg installs will fail (no way to work around that), rpm installs won't fail as long as the Agent is pinned to a version that's signed with the older keys.

AFAIK we import the new keys safely (through a keyserver with `apt`, and with a fingerprint check with `rpm`).

More details in the commit messages.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -91,6 +91,10 @@ default['datadog']['yumrepo_proxy_username'] = nil
 default['datadog']['yumrepo_proxy_password'] = nil
 default['datadog']['windows_agent_url'] = 'https://s3.amazonaws.com/ddagent-windows-stable/'
 
+# Location of additional rpm gpgkey to import (with signature `e09422b3`). In the future the rpm packages
+# of the Agent will be signed with this key.
+default['datadog']['yumrepo_gpgkey_new'] = "#{yum_protocol}://yum.datadoghq.com/DATADOG_RPM_KEY_E09422B3.public"
+
 # Agent installer checksum
 # Expected checksum to validate correct agent installer is downloaded (Windows only)
 default['datadog']['windows_agent_checksum'] = nil
--- a/recipes/repository.rb
+++ b/recipes/repository.rb
@@ -25,6 +25,13 @@ when 'debian'
     action :install
   end
 
+  # Trust new APT key
+  execute 'apt-key import key 382E94DE' do
+    command 'apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 A2923DFF56EDA6E76E55E492D3A80E30382E94DE'
+    not_if 'apt-key list | grep 382E94DE'
+  end
+
+  # Add APT repository
   apt_repository 'datadog' do
     keyserver 'hkp://keyserver.ubuntu.com:80'
     key 'C7A7DA52'
--- a/recipes/repository.rb
+++ b/recipes/repository.rb
@@ -44,6 +44,30 @@ when 'debian'
 when 'rhel', 'fedora'
   include_recipe 'yum'
 
+  # Import new RPM key
+  if node['datadog']['yumrepo_gpgkey_new']
+    # gnupg is required to check the downloaded key's fingerprint
+    package 'gnupg' do
+      action :install
+    end
+
+    # Download new RPM key
+    key_local_path = ::File.join(Chef::Config[:file_cache_path], 'DATADOG_RPM_KEY_E09422B3.public')
+    remote_file 'DATADOG_RPM_KEY_E09422B3.public' do
+      path key_local_path
+      source node['datadog']['yumrepo_gpgkey_new']
+    end
+
+    # Import key if fingerprint matches
+    execute 'rpm-import datadog key e09422b3' do
+      command "rpm --import #{key_local_path}"
+      not_if 'rpm -q gpg-pubkey-e09422b3' # (key already imported)
+      only_if "gpg --dry-run --quiet --with-fingerprint #{key_local_path} | grep 'A4C0 B90D 7443 CF6E 4E8A  A341 F106 8E14 E094 22B3'"
+      action :run
+    end
+  end
+
+  # Add YUM repository
   yum_repository 'datadog' do
     name 'datadog'
     description 'datadog'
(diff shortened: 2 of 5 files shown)


attributes/default.rb
```diff
@@ -91,6 +91,10 @@
 default['datadog']['yumrepo_proxy_password'] = nil
 default['datadog']['windows_agent_url'] = 'https://s3.amazonaws.com/ddagent-windows-stable/'
 
+# Location of additional rpm gpgkey to import (with signature `e09422b3`). In the future the rpm packages
+# of the Agent will be signed with this key.
+default['datadog']['yumrepo_gpgkey_new'] = "#{yum_protocol}://yum.datadoghq.com/DATADOG_RPM_KEY.public.E09422B3"
```
olivielpeau: @yannmh: the new rpm key currently lives at this URL, I'm not a huge fan of its name though, I'd prefer something like `DATADOG_RPM_KEY_E09422B3.public`, what do you think?

olivielpeau: Also: we can discuss the new attribute name, I dislike `new` in an attribute name but haven't found a better alternative

recipes/repository.rb
```diff
@@ -37,6 +44,30 @@
...
   include_recipe 'yum'
 
+  # Import new RPM key
+  if node['datadog']['yumrepo_gpgkey_new']
+    # gnupg is required to check the downloaded key's fingerprint
+    package 'gnupg' do
+      action :install
+    end
+
+    # Download new RPM key
```
truthbk: I don't think you need to download the file... You should be able to `rpm --import "#{yum_protocol}://yum.datadoghq.com/DATADOG_RPM_KEY.public.E09422B3"`

olivielpeau: Actually I did this to make sure that we're trusting the right key (in particular we download the key through plain http on CentOS 5 so I think it's better to check that we are importing the correct key). What do you think?

truthbk: Makes sense, should've made the connection with: https://github.com/DataDog/chef-datadog/pull/365/files#diff-62b784dc2245888950b940c524931921R65

yannmh:
### Installation/upgrades scenario

**Datadog Agent signed with the _old_ key**
```
+======================+============+==============+
| Scenario\Cookbook    | < versions | this version |
+----------------------+------------+--------------+
| New install          | ✓          | ✓            |
+----------------------+------------+--------------+
| Upgrade from > 5.8.5 | ✓          | ✓            |
+----------------------+------------+--------------+
| Upgrade from < 5.8.5 | ✓          | ✓            |
+----------------------+------------+--------------+
```


**Datadog Agent signed with the _new_ key**
```
+======================+============+==============+
| Scenario\Cookbook    | < versions | this version |
+----------------------+------------+--------------+
| New install          | ✗          | ✓            |
+----------------------+------------+--------------+
| Upgrade from > 5.8.5 | ✓          | ✓            |
+----------------------+------------+--------------+
| Upgrade from < 5.8.5 | ✗          | ✓            |
+----------------------+------------+--------------+
```

recipes/repository.rb
```diff
@@ -25,6 +25,13 @@
     action :install
   end
 
+  # Trust new APT key
+  execute 'install new apt key' do
+    command 'apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 A2923DFF56EDA6E76E55E492D3A80E30382E94DE'
+    not_if 'apt-key list | grep 382E94DE'
+  end
+
```
yannmh: Naive question: can we simply use the `apt` cookbook twice to do that, i.e.
```ruby
# Legacy key
  apt_repository 'datadog' do
    keyserver 'hkp://keyserver.ubuntu.com:80'
    key 'C7A7DA52'
    uri node['datadog']['aptrepo']
    distribution node['datadog']['aptrepo_dist']
    components ['main']
    action :add
  end

  # New key
  apt_repository 'datadog' do
    keyserver 'hkp://keyserver.ubuntu.com:80'
    key '382E94DE'
    uri node['datadog']['aptrepo']
    distribution node['datadog']['aptrepo_dist']
    components ['main']
    action :add
  end
```

olivielpeau: Hmm we could use the resource twice (as long as we name the 2 differently, otherwise we're cloning the resource, which is a Bad thing to do), but that would define 2 `.list` files with the same repository, which `apt` doesn't like (it will spew out warnings because the repo is duplicated).

And it wouldn't look good either. That said I'll look into it a bit more to see if we couldn't make use of the logic that's already implemented in the `apt` cookbook.

yannmh: No need to spend too much time on this. 

I am just a little worried about the "flow" for trusting the new RPM key. It involves multiple commands, with `|` and `grep`. I was wondering if we can be sure that the output will be consistent across all systems using RPM and the different versions of `rpm-import`.

olivielpeau: So I had another look at the `apt` cookbook, and unfortunately we can't really re-use anything from there, so I think I have to stick to the current approach.

About the rpm key: from what I've tested the `rpm --import`, `rpm -q` and `gpg --with-fingerprint <keyfile>` have very consistent behaviors (I've checked manually on CentOS 5 and 7). I understand your concern though, especially on the `grep` command which could be very sensitive to the slightest change in `gpg`'s output, I'll see if this could be made a little clearer/more resilient.

olivielpeau:
Added spec and integration tests, @yannmh could you have one last look? 🙇 

Also, do you have an opinion about the name of the new RPM key? (both its current URL and the attribute name) If you're fine with them then we're all set

attributes/default.rb
yannmh: > I'd prefer something like DATADOG_RPM_KEY_E09422B3.public, what do you think?

👍

olivielpeau: thanks, updated to `DATADOG_RPM_KEY_E09422B3.public`

PR #368, opened 2016-10-06
Fixes #366 - Add ability to pin datadog-agent versions to install for multiple platforms

### LINKED ISSUE #366: Attribute for version of datadog-agent to install is the same for Linux and Windows install packages
The attribute for version of the datadog-agent to be installed is the same for Linux and Windows install packages, and the problem with that is that the version number format of these packages does not match.  For example, the latest package for windows is `ddagent-cli-5.9.0.msi` and the latest for linux is `datadog-agent-5.9.0-1.x86_64.rpm`.  Linux has a `-1` and Windows does not (as is standard).

So if we were to pin a version with `node['datadog']['agent_version']` at `'5.9.0-1'` it would work for Linux, not Windows, and vice versa if we pinned it to `'5.9.0'`.  If we want to use the same attribute we have to assign the proper value of that attribute based on the platform of the node in a wrapper recipe; we couldn't just use a role/chef_environment since one platform or the other wouldn't like the value of the version.

The easiest thing might be to have
```
default['datadog']['linux_agent_version']
default['datadog']['win_agent_version']
```
and then you can set each of these separately.

mlcooper:


iancward:
@olivielpeau since Fedora and RHEL pull from the same repository, we figured it would be better to have both platform families use the same attribute key for the agent version. However, we could just as easily remove that logic, which would require a user to set a separate `fedora` and `rhel` attribute.
Let @mlcooper and I know which way makes the most sense.

spec/dd-agent_spec.rb
```diff
@@ -205,6 +205,122 @@ def set_env_var(name, value)
...
 
+  context 'allows a string for agent version' do
+    cached(:chef_run) do
+      ChefSpec::SoloRunner.new(
+        :platform => 'ubuntu',
+        :version => '14.10'
+      ) do |node|
+        node.set['datadog'] = {
+          'api_key' => 'somethingnotnil',
+          'agent_version' => '4.4.0-200'
```
olivielpeau: nit: could you use a version with an epoch here since the platform is ubuntu? (`1:5.9.0-1`)

spec/dd-agent_spec.rb
```diff
@@ -205,6 +205,122 @@ def set_env_var(name, value)
...
+        expect(chef_run).to install_apt_package('datadog-agent').with(version: '1:5.9.0-1')
+      end
+    end
+
+    context 'when windows' do
+      cached(:chef_run) do
+        ChefSpec::SoloRunner.new(
+          :platform => 'windows',
+          :version => '2012R2',
+          :file_cache_path => '/tmp'
```
olivielpeau: I'd prefer a "realistic" path on Windows (`C:/chef/cache` for instance, see https://github.com/DataDog/chef-datadog/blob/v2.6.0/spec/dd-agent_spec.rb#L162)

spec/dd-agent_spec.rb
```diff
@@ -205,6 +205,122 @@ def set_env_var(name, value)
...
+
+      temp_file = ::File.join('/tmp', 'ddagent-cli.msi')
+
+      # remote_file source gets converted to an array, so we need to do
+      # some tricky things to be able to regex against it
+      # Relevant: http://stackoverflow.com/a/12325983
+      # But we should probably assert the full default attribute somewhere...
+      it 'installs agent 4.4.0' do
+        expect(chef_run.remote_file(temp_file).source.to_s)
+          .to match(/ddagent-cli-4.4.0.msi/)
```
olivielpeau: 👍

spec/dd-agent_spec.rb
```diff
@@ -205,6 +205,122 @@ def set_env_var(name, value)
...
+              'debian' => '1:5.9.0-1',
+              'rhel' => '4.4.0-200',
+              'windows' => '4.4.0'
+            }
+          }
+        end.converge described_recipe
+      end
+
+      temp_file = ::File.join('/tmp', 'ddagent-cli.msi')
+
```
olivielpeau: could you add the common example `it_behaves_like 'windows Datadog Agent'` here?

(you'll also need to set the env var `ProgramData` like here: https://github.com/DataDog/chef-datadog/blob/v2.6.0/spec/dd-agent_spec.rb#L158)

mlcooper:
Hi @olivielpeau please let us know if you're okay with the changes made per your feedback.  Thanks!

olivielpeau:
Looks great, thanks @mlcooper and @iancward!

Merging

PR #374, opened 2016-10-17
Fixed #373 - Postfix template issue

### LINKED ISSUE #373: Postfix template adds tags line even if no tags exist.
chef-datadog/templates/default/postfix.yaml.erb (current):

```
    tags:
      <% Array(instance['tags']).each do |tag| -%>
      - <%= tag %>
      <% end -%>
```

This code segment adds a line with 'tags:' even if there is no 'tags' element. In the dd-agent code, a tags line like this ends up being interpreted as `None` (correctly) instead of `[]` (empty array). This causes the integration to fail to send data correctly, even though 'tags' is listed as optional.

Suggestion:
Check for presence of 'tags' before adding 'tags' line (as seen in other templates, such as https://github.com/DataDog/chef-datadog/blob/master/templates/default/lighttpd.yaml.erb).

nyanshak:


olivielpeau:
Thanks @nyanshak. I've added a spec test, this looks ready for the next minor release.

Merging

Issue #336, opened 2016-08-05
allow custom configuration in datadog.conf


degemer:
For now we can use the `legacy_integrations`: https://github.com/DataDog/chef-datadog/blob/fdc80b464c0539be2720f0c6e0626b6d2e64ead1/templates/default/datadog.conf.erb#L116-L135

But it's clearly not a good way to do it, it would be way easier if we were able to add custom 'key: value' with an attribute like `node['datadog']['extra_config'] = {key: value}`.

PR #375, opened 2016-10-20
Add support for any config options

degemer:
It is annoying to have to wait for a new cookbook release when an agent
is released with a new config option.
This commit aims to fix it by allowing any config options to be added.

Fix https://github.com/DataDog/chef-datadog/issues/336

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -218,6 +218,11 @@ default['datadog']['statsd_metric_namespace'] = nil
 default['datadog']['histogram_aggregates'] = 'max, median, avg, count'
 default['datadog']['histogram_percentiles'] = '0.95'
 
+# extra config options
+# If an agent is released with a new config option which is not yet supported by this cookbook
+# you can use this attribute to set it. Will be ignored if nil.
+default['datadog']['extra_config']['forwarder_timeout'] = nil
+
 # extra_packages to install
 default['datadog']['extra_packages'] = {}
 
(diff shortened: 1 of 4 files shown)


attributes/default.rb
```diff
@@ -221,6 +221,11 @@
 # extra_packages to install
 default['datadog']['extra_packages'] = {}
 
+# extra config options
+# If an agent is released with a new config option which is not yet supported by this cookbook
+# you can use this attribute to set it. Will be ignored if falsy.
```
olivielpeau: I think we should ignore the value _only if_ it's `nil`. We have agent configuration options that are `true` by default so setting them to `false` is meaningful. Thoughts?

attributes/default.rb
```diff
@@ -221,6 +221,11 @@
 # extra_packages to install
 default['datadog']['extra_packages'] = {}
 
+# extra config options
```
olivielpeau: nit: could you put this right under `default['datadog']['histogram_percentiles']` in this file, so that it's visually regrouped with the other agent config attribute defaults?

attributes/default.rb
olivielpeau: (also, if you agree, a test on that would be perfect :) )

degemer: Makes sense. I didn't think of `false`, only of `''`, which is useless since ini requires key: value.

PR #380, opened 2016-11-13
Simplify postgres.yaml template

### LINKED ISSUE #379: postgres: template file renders invalid file in some cases
Hello,

I've run in to some issues with the template file used to render the `check.d` configuration for Postgres. Specifically, the file is unpacking the attributes (by hand) in to YAML. Unfortunately, there are a few cases where this file is doing the wrong thing.

The first issue I ran in to is that when using the `relations` attribute (Array), the template expects each entry to be a string[1]. This is at odds with the spec because a relation can be defined like so:

```YAML
relations:
  - relation_name: table
    schemas:
      - a
      - b
```

Trying to use an Array of Hashes gives me:

```YAML
relations:
  - {"relation_name"=>"table","schemas"=>["a","b"]}
```

The second issue is that the template doesn't allow the `descriptors` to be an empty array[2], and omitting the key is a error according to `dd-agent info`. If you try to provide an empty Array for descriptors, and assuming `relation` is false, it renders the file as such which triggers a parsing error:

```YAML
descriptors:
relation: false
```

I'd expect it to render like this:

```YAML
descriptors: []
relation: false
```

My suggested fix would be to change `templates/default/postgres.yaml.erb` to use `.to_yaml` to render the specific keys and avoid trying to manually unpack attributes in to YAML:

```Ruby
# .to_a is used to convert the attributes so that they don't marshal custom tags
<%= { "init_config" => {}, "instances" => @instances.to_a }.to_yaml %>
```

Unfortunately, the template that's being rendered does some things that alter attributes to provide the correct key name[3] or provide a default value[4]. This means changing this file to use `.to_yaml` would introduce breaking changes.

To mitigate one of these concerns, one option would be that we can patch `dd-agent` to default to port `5432`[5] (as well as default to localhost while we're there too).

That still leaves the `host` => `server` mismatch[3] as a breaking change. To mitigate that, we _could_ update the `postgres` recipe to iterate over the instances and munge the values to either set defaults or map the names to the right values. But this feels gross to me.

What do you think? In theory, what would be needed to release a breaking change and how long would it take to see a release contain the changes?

[1] [github.com/DataDog/chef-datadog/templates/default/postgres.yaml.erb#L23-L26](https://github.com/DataDog/chef-datadog/blob/32bef19902ce1b547321d701bc590ca34127ef5c/templates/default/postgres.yaml.erb#L23-L26)
[2] [github.com/DataDog/chef-datadog/templates/default/postgres.yaml.erb#L43-L46](https://github.com/DataDog/chef-datadog/blob/32bef19902ce1b547321d701bc590ca34127ef5c/templates/default/postgres.yaml.erb#L43-L46)
[3] [github.com/DataDog/chef-datadog/templates/default/postgres.yaml.erb#L6](https://github.com/DataDog/chef-datadog/blob/32bef19902ce1b547321d701bc590ca34127ef5c/templates/default/postgres.yaml.erb#L6)
[4] [github.com/DataDog/chef-datadog/templates/default/postgres.yaml.erb#L7](https://github.com/DataDog/chef-datadog/blob/32bef19902ce1b547321d701bc590ca34127ef5c/templates/default/postgres.yaml.erb#L7)
[5] [github.com/DataDog/dd-agent/checks.d/postgres.py#L661](https://github.com/DataDog/dd-agent/blob/d3421104af5d511acbb3f8be37d9c1506d4dabab/checks.d/postgres.py#L661)

miketheman:
Converts the postgres template to using the "simpler" rendering
solution, instead of manually unpacking positional values in YAML.

Using in-template replacement logic, we can modify the values in-place
at the time of rendering.
This does increase memory overhead slightly, as the `instances` object
is copied once, as it starts out as a `Chef::ImmutableArray` that prevents
modification.

- Adds spec tests for specific uses brought up in #379.
- Updates in-recpie documentation to show that one can use either
`server` or `host` values
- Removes a breaking change TODO

Resolves #379

Signed-off-by: Mike Fiedler <miketheman@gmail.com>


miketheman:
@theckman Care to try this branch?

theckman:
@miketheman I'll give it a go this afternoon. If you're curious, ere is the branch I have WIP that seemed to have fixed my issues (minus the breaking changes I discussed in #379): https://github.com/theckman/chef-datadog/tree/fix_postgres_yaml -- I was going to get around to cleaning it up today, but maybe now I can be lazy.

templates/default/postgres.yaml.erb
```diff
@@ -1,53 +1,21 @@
-init_config:
-# empty by design
+# Generated by Chef, local modifications will be overwritten
+
+<%# TODO: Breaking change, remove defaults -%>
```
theckman: It's a bummer the `datadog_monitor` LWRP doesn't provide a mechanism for passing the raw file content in to the LWRP. Over time I've tried to avoid doing logic like this in the template file, because they can get pretty intense pretty quick.

I wonder if there is any interest in supporting that within the LWRP itself.

templates/default/postgres.yaml.erb
```diff
@@ -1,53 +1,21 @@
...
-        relation: <%= cm["relation"] %>
-        <% end -%>
-      <% end -%>
-    <% end -%>
-<% end -%>
+    # If `port` is unspecified, set to 5432
+    instance['port'] = 5432 if instance['port'].nil?
+  end
+-%>
+<%= JSON.parse(({'instances' => instances}).to_json).to_yaml %>
```
theckman: Nice hack for avoiding there being a second document (`---`) in this file. Hadn't even thought about putting that before the `init_config` line and was instead just building a Hash in the template file and rendering that to YAML.

theckman:
Actually, there may be one breaking change in here around the `relations` key. I ran in to this issue as I converted more things to use the updated version.

Because each item in the `relations` key was treated as a string before, and it was supposed to be a hash, I ended up with the following block of code in my wrapper recipe:

```Ruby
tables = %w(a b c)
relations = tables.map { |t| "relation_name: #{t}" }
```

This is the only way I could get it to render like so:

```YAML
relations:
  - relation_name: a
  - relation_name: b
  - relation_name: c
```

With this change the file rendered like:

```YAML
relations:
  - 'relation_name: a'
  - 'relation_name: b'
  - 'relation_name: c'
```

miketheman:
@theckman is there an issue with backing out your wrapper recipe changes? I don't think the string vs has was ever clearly distinguished and as you've mentioned, you had to be creative in working around it, so to me that points at a bug in the approach.

theckman:
@miketheman I'm all :+1: for introducing this change and fixing things on my side. I had some concerns that others may have done something similar, and wanted to call it out as a consideration.  I've only did it in one place so far since I'm just starting to roll out Datadog, so it's easy for me.

PR #388, opened 2017-01-03
Rename package[apt-transport-https] resource for Chef 13 compatibility

bai:
This PR renames `package[apt-transport-https]` to avoid resource cloning.

Currently, when DataDog cookbook used with `apt` cookbook 2.9.1 or later (released October 24, 2015), Chef throws a deprecation warning:

```
Deprecated features used!
  Cloning resource attributes for apt_package[apt-transport-https] from prior resource
Previous apt_package[apt-transport-https]: /var/chef/cache/cookbooks/apt/recipes/default.rb:85:in `from_file'
Current  apt_package[apt-transport-https]: /var/chef/cache/cookbooks/datadog/recipes/repository.rb:24:in `from_file' at 1 location:
    - /var/chef/cache/cookbooks/datadog/recipes/repository.rb:24:in `from_file'
   See https://docs.chef.io/deprecations_resource_cloning.html for further details.
```

Originally this PR suggested to just remove the resource and bump `apt` cookbook dependency but it's been decided to rename the resource instead — for those who depend on an older `apt` cookbook elsewhere.

### DIFF
--- a/recipes/repository.rb
+++ b/recipes/repository.rb
@@ -21,7 +21,8 @@ case node['platform_family']
 when 'debian'
   include_recipe 'apt'
 
-  package 'apt-transport-https' do
+  package 'install-apt-transport-https' do
+    package_name 'apt-transport-https'
     action :install
   end
 


degemer:
Hello @bai, thank you for your contribution and explanation !

We try to avoid as much as possible setting requirements for our dependencies if there is another way around (even if the requirement is loose), and I think there is be in this case: renaming the resource. If I'm right, this would allow us to maintain our current level of compatibility with the `apt` cookbook while being compatible with Chef 13.

On a side note, do you know when Chef 13 will be released ? I couldn't find an official release date.

In the long term, we will definitely use your solution, but we know that for now some of our customers are still using an older `apt` version, and don't want to force them to upgrade.

Let me know what you think or our proposed solution, or if you have any question!

martinisoft:
> On a side note, do you know when Chef 13 will be released ? I couldn't find an official release date.

This will happen April 2017 per [this accepted RFC for release cadence](https://github.com/chef/chef-rfc/blob/master/rfc081-release-cadence.md#chef-release-cadence)

bai:
Apologies with getting this one forgotten, I've pushed the fix. 👀

bai:
Friendly bump 💯 

cc: @olivielpeau

andrewjamesbrown:
@degemer I noticed that there are also Chef13 deprecation warnings/errors related to https://github.com/DataDog/chef-datadog/blob/master/recipes/dd-handler.rb#L90

See https://github.com/chef-cookbooks/chef_handler/issues/45 for more details

degemer:
Thanks for reporting this @andrewjamesbrown !

Do you have the same issue with chef_handler 2.x ? We haven't had a chance to test it yet, but if it fixes this issue and the behaviour stays the same, we could relax this requirement: https://github.com/DataDog/chef-datadog/blob/master/metadata.rb#L25

PR #389, opened 2017-01-03
[rubocop] Fix rubocop offenses

olivielpeau:
Also adds an exclusion for the block length cop on our `test/` folder.

No functional change.

Closes #356 (ended up choosing the `644` notation for the file/dir `mode`s)

### DIFF
--- a/recipes/dd-handler.rb
+++ b/recipes/dd-handler.rb
@@ -73,4 +73,5 @@ chef_handler 'Chef::Handler::Datadog' do
   arguments [handler_config]
   supports :report => true, :exception => true
   action :nothing
-end.run_action(:enable) if node['datadog']['chef_handler_enable']
+  only_if { node['datadog']['chef_handler_enable'] }
+end.run_action(:enable)
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,6 +10,15 @@ AllCops:
 Metrics/LineLength:
   Max: 160
 
+# Exclude Rakefile and tests since there are long blocks that aren't worth shortening there
+# Also, slightly longer blocks are fine
+Metrics/BlockLength:
+  Max: 30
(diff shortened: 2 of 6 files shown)


recipes/dd-handler.rb
```diff
@@ -73,4 +73,5 @@
   arguments [handler_config]
   supports :report => true, :exception => true
   action :nothing
-end.run_action(:enable) if node['datadog']['chef_handler_enable']
+  only_if { node['datadog']['chef_handler_enable'] }
```
miketheman: Need to verify if the `only_if` is evaluated at compile time.

olivielpeau: Thanks for pointing this out, I just verified that the guard is evaluated correctly at compile time, on both chef 11 and 12.

i.e with `node['datadog']['chef_handler_enable']` set to `false`, running `datadog:dd-handler` outputs the following:

```
Compiling Cookbooks...
[...]
Recipe: datadog::dd-handler
  * chef_gem[chef-handler-datadog] action install (up to date)
  * chef_handler[Chef::Handler::Datadog] action enable (skipped due to only_if)

Converging 3 resources
[...]
Recipe: datadog::dd-handler
  * chef_gem[chef-handler-datadog] action install (up to date)
  * chef_handler[Chef::Handler::Datadog] action nothing (skipped due to action :nothing)
```

miketheman: Nice! Worthwhile adding a spec in https://github.com/DataDog/chef-datadog/blob/master/spec/dd-handler_spec.rb to verify the negative (i.e does not load the handler)?

olivielpeau: 👍  Added a spec

olivielpeau:
Thank you both for the reviews! 🙇

Issue #369, opened 2016-10-07
How can I use the datadog_monitor resource to install my own custom checks?


red8888:
I'm looking at the monitoring resource and I'm curious how I can use it to install my own checks.

I can see how the **:add** action copies down the .yaml files from the templates folder but where does it copy the .py files for the checks?

olivielpeau:
Hi @red8888!

The `datadog_monitor` resource doesn't support custom checks right now, but it'd make sense to make it support it. Feel free to open a PR implementing this, if you can't I'm adding this issue to our triage along with the related issue #309 so that we take care of it in the future.

Let me know if you have any questions!

Issue #309, opened 2016-05-19
provide resource to roll-out custom AgentChecks


rmoriz:
As a DataDog user
I've written a custom AgentCheck
and want to use chef-datadog
to deploy the check and its config to my nodes

olivielpeau:
Hi @rmoriz,

As you wrote, this could mainly addressed through some improved documentation.

Just to be sure I understand, could you give an example of what you mean by "custom template source definition"? Thanks!

rmoriz:
1. The current `monitor` resource only manages .yml files in `/etc/dd-agent/conf.d/`

Let's say you have authored a custom check, you'll have to poke into the "internals" and manually place the check.

```ruby
template '/etc/dd-agent/checks.d/systemd_check.py' do
  source 'systemd_check.py.erb'
end
```

2. custom template source

```ruby 
datadog_monitor 'systemd_check'
```

while `datadog_monitor` provides the option to set [a source cookbook](https://github.com/DataDog/chef-datadog/blob/fdc80b464c0539be2720f0c6e0626b6d2e64ead1/providers/monitor.rb#L26), there is no way to overwrite the `source` attribute of the template resource.

olivielpeau:
Thanks @rmoriz for detailing your use case!

`1.` could be addressed through some improved documentation. If the use case is common enough we could also consider adding a `dd-custom-check` recipe that takes care of installing the custom check file and calls the `monitor` resource.

For `2.` I agree that it could make sense to allow setting a custom source template on the resource.

Adding both to our triage so that we take care of it at some point, but feel free to open PRs if you have the opportunity!

PR #391, opened 2017-01-05
Support custom checks

olivielpeau:
**WIP, open for discussion**

- Update of the LWRP: for custom checks, we copy the python file first and then proceed with the yaml file
- `custom_checks` recipe
- spec tests

The custom python checks live in other cookbooks (the cookbook from which the custom check file is pulled can be configured either as an attribute or directly on the `datadog_monitor` resource).

Only the generic yaml template (`integrations.erb`) can be used.

If this approach looks good I can go ahead and implement it for windows too.

Should address #309 and #369

TODO:
- support of `:remove` action
- windows support (could be done later)

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,3 +112,18 @@ Usage
 We are not making use of data_bags in this recipe at this time, as it is unlikely that you will have more than one API key and one application key.
 
 For more deployment details, visit the [Datadog Documentation site](http://docs.datadoghq.com/).
+
+Custom checks
+-------------
+To install your own custom checks with this cookbook, please follow the examples in the related [recipe](https://github.com/DataDog/chef-datadog/tree/master/recipes/custom_checks.rb).
+The python check needs to live as a cookbook file in one of your cookbooks.
+
+Alternatively, you can use the `datadog_monitor` resource from your own cookbooks:
+
+```ruby
+datadog_monitor 'my_custom_check' do
+  init_config {}
+  instances [{'parameter1': 'value1'}]
+  custom_check true
+end
+```
--- a/providers/monitor.rb
+++ b/providers/monitor.rb
@@ -7,8 +7,19 @@ def whyrun_supported?
   true
 end
 
+def install_custom_check(name, cookbook, source)
+  cookbook_file ::File.join(node['datadog']['config_dir'], 'checks.d', "#{name}.py") do
+    cookbook cookbook unless cookbook.nil?
+    source source unless source.nil?
+    owner 'dd-agent'
+    mode '644'
+  end
+end
+
 action :add do
   Chef::Log.debug "Adding monitoring for #{new_resource.name}"
+  install_custom_check(new_resource.name, new_resource.check_cookbook, new_resource.check_source) if new_resource.custom_check
+
   template ::File.join(node['datadog']['config_dir'], 'conf.d', "#{new_resource.name}.yaml") do
     if node['platform_family'] == 'windows'
       owner 'Administrators'
@@ -19,7 +30,7 @@ action :add do
       mode '600'
     end
 
-    source 'integration.yaml.erb' if new_resource.use_integration_template
+    source 'integration.yaml.erb' if new_resource.use_integration_template || new_resource.custom_check
 
     variables(
       init_config: new_resource.init_config,
--- a/recipes/custom_checks.rb
+++ b/recipes/custom_checks.rb
@@ -1,0 +1,60 @@
+#
+# Cookbook Name:: datadog
+# Recipe:: custom_checks
+#
+# Copyright 2011-2017, Datadog
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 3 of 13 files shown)


README.md
```diff
@@ -112,3 +112,18 @@ Usage
...
+To install your own custom checks with this cookbook, please follow the examples in the related [recipe](https://github.com/DataDog/chef-datadog/tree/master/recipes/custom_checks.rb).
+The python check needs to live as a cookbook file in one of your cookbooks.
+
+Alternatively, you can use the `datadog_monitor` resource from your own cookbooks:
+
+```ruby
+datadog_monitor 'my_custom_check' do
+  init_config {}
+  instances [{'parameter1': 'value1'}]
+  custom_check true
```
degemer: I'd prefer the opposite (i.e. default `true`), and we would set it to false in our own recipes (or even better, remove it and determine it based on the known integrations), but this would be a breaking change. What do you think of doing this for the next major ?

providers/monitor.rb
```diff
@@ -7,8 +7,19 @@ def whyrun_supported?
   true
 end
 
+def install_custom_check(name, cookbook, source)
+  cookbook_file ::File.join(node['datadog']['config_dir'], 'checks.d', "#{name}.py") do
+    cookbook cookbook unless cookbook.nil?
+    source source unless source.nil?
+    owner 'dd-agent'
+    mode '644'
```
degemer: People might prefer a stricter mode for their own checks. `600` ?  ¯\_(ツ)_/¯

README.md
degemer: I think we should use the `check_cookbook` arg, since people will have to specify it if they want to use `datadog_monitor` in their own cookbooks.

providers/monitor.rb
```diff
@@ -7,8 +7,19 @@ def whyrun_supported?
   true
 end
 
+def install_custom_check(name, cookbook, source)
+  cookbook_file ::File.join(node['datadog']['config_dir'], 'checks.d', "#{name}.py") do
```
degemer: Nice default!

recipes/custom_checks.rb
```diff
@@ -0,0 +1,60 @@
...
+#   'default_timeout' => 10
+# }
+# default['datadog']['my_custom_check_2']['instances'] = [
+#   {
+#     'url' => 'http://localhost:22222'
+#   }
+# ]
+node['datadog']['custom_checks'].each do |check_name, custom_check|
+  next unless custom_check['enabled']
+
```
degemer: Should we fail there with an explicit message if `node['datadog'][check_name]['init_config']` or `node['datadog'][check_name]['instances']` is missing? I think we could have a better message than a attribute missing error, what do you think?

recipes/dd-agent.rb
```diff
@@ -103,3 +103,6 @@
 
 # Install integration packages
 include_recipe 'datadog::integrations' unless is_windows
+
+# Install custom checks
+include_recipe 'datadog::custom_checks' unless is_windows # FIXME: implement this on windows?
```
degemer: I think Windows support could be as easy adding a `if` on `cookbook_file` args
```ruby
owner 'dd-agent'
mode '644'
```
and use the Windows equivalent in this case. Anyway, we can target the next release.

spec/spec_helper.rb
```diff
@@ -9,6 +9,8 @@
 require 'shared_examples'
 
 RSpec.configure do |config|
+  config.berkshelf_options = { except: :integration }
```
degemer: It's not needed anymore? We can delete it in the Berksfile directly then, no?

README.md
olivielpeau: Why do you think having a default of `true` makes more sense? If I had to guess I'd say that most of the time `datadog_monitor` is used for non-custom checks

olivielpeau: ok so here I left the default of `check_cookbook` to `nil` so that by default when no `check_cookbook` is specified the python check will be searched in the files of the cookbook that used the `datadog_monitor` resource.

So if people use `datadog_monitor` from their own cookbooks they generally shouldn't have to specify the `check_cookbook`. But if they use the `datadog::custom_checks` checks recipe they'll have to specify it as an attribute (otherwise chef will try to find the python check in the files of the `datadog` cookbook and wouldn't find it obviously)

recipes/custom_checks.rb
olivielpeau: yes, good point 👍

spec/spec_helper.rb
olivielpeau: I _think_ it's still needed for the kitchen tests, I'll have to check that though

olivielpeau:
Going to move this back to a later release, the PR is not ready enough to be shipped with `2.8.0`. We should also discuss if it would be better to ship this with a major version bump.

README.md
degemer: I think it would make it easier for people to use it for their checks. And it doesn't cost us a lot to modify our checks not to use the default. I don't think it's really important though.

AliDatadog:
Closed for inactivity.

PR #394, opened 2017-01-19
Add mesos_slave/mesos_master templates and recipes

Carles-Figuerola:
Add support to enable the mesos_slave.yaml file found in datadog's conf folder from the chef cookbook


recipes/mesos_slave.rb
```diff
@@ -0,0 +1,33 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+include_recipe 'datadog::dd-agent'
+
+# Build a data structure with configuration.
+# @example
+#   node.override['datadog']['mesos'] =
```
olivielpeau: for consistency the key should be `mesos_slave`

recipes/mesos_slave.rb
```diff
@@ -0,0 +1,33 @@
...
+#     instances: [{
+#       url: 'http://server:port',
+#       timeout: 8
+#     }],
+#     init_config: {
+#       default_timeout: 10
+#     }
+
+datadog_monitor 'mesos_slave' do
```
olivielpeau: You need to specify the `init_config` and `instances` properties of the resource, with `node['datadog']['mesos_slave']['init_config']` and `node['datadog']['mesos_slave']['instances']` as values

templates/default/mesos_slave.yaml.erb
```diff
@@ -0,0 +1,13 @@
+# Generated by Chef, local modifications will be overwritten
```
olivielpeau: Any reason not to use a generic template, i.e. use the approach of this template https://github.com/DataDog/chef-datadog/blob/v2.7.0/templates/default/docker_daemon.yaml.erb?

recipes/mesos_slave.rb
Carles-Figuerola: true, changed

recipes/mesos_slave.rb
Carles-Figuerola: added, but allowed optionally using them or not

templates/default/mesos_slave.yaml.erb
Carles-Figuerola: A custom template was used to allow the defaults to be used without having to define extra attributes

templates/default/mesos_master.yaml.erb
```diff
@@ -0,0 +1,13 @@
+# Generated by Chef, local modifications will be overwritten
+
+init_config:
+  <% if @init_config.key?('default_timeout') -%>
+  default_timeout: <%= @init_config['default_timeout'] %>
+  <% end -%>
+
+instances:
+  <% if @init_config.key?('instances') -%>
```
olivielpeau: The condition here doesn't look correct (if there's something to check it should be on the `@instances` variable directly, `@init_config` should never have an `instances` key).

More generally, what you could instead here would be to format the `@instances` array to yaml (and, if no `@instances` have been passed, i.e. when `@instances` is an empty array, just define an instance with default values). It would also allow users to set the other configuration options (for example `disable_ssl_validation` for `mesos_master`, see https://github.com/DataDog/dd-agent/blob/5.10.1/conf.d/mesos_master.yaml.example#L15).

So something like: (haven't checked that the syntax is correct but just to describe the logic I'm thinking of)

```erb
instances:
<% if @instances.empty? -%>
- url: "http://<%= @node['ipaddress'] %>:5050"
<% else %>
<%= JSON.parse((@instances).to_json).to_yaml %>
<% end -%>
```

This also applies to `mesos_slave` ;)

templates/default/mesos_master.yaml.erb
```diff
@@ -0,0 +1,13 @@
...
+  - url: <%= @init_config['instances'] %>
+  <% else %>
+  - url: "http://<%= @node['ipaddress'] %>:5050"
```
olivielpeau: why use `@node['ipaddress']` instead of `localhost` here?

templates/default/mesos_slave.yaml.erb
olivielpeau: ok, thanks for the clarification! This is 👍 to me right now.

Just so you know, in a future major version of this cookbook we'll move towards generalizing the installation of integrations (to simplify the cookbook and make it more flexible) so we'll probably stop supporting integration-specific yaml templates.

AliDatadog:
Closed for inactivity.

Issue #377, opened 2016-11-01
Add support for retrieving secrets from chef-vault


theckman:
With newer versions of `chef-client`, it's nice to work with `chef-vault` to manage secrets for system utilities. As of right now, the only mechanism for using `chef-vault` and the `datadog` cookbook is to write a wrapper cookbook, and set the attributes. That works great. The only downside is that we're setting attributes, and thus effectively leaking credentials.

Is there any interest in enhancing the installation codepath to conditionally use `chef-vault` for pulling credentials and setting them in the configuration files? That way we could avoid leaking plain text credentials in to the node object, while still keeping the old behavior as the default (for now)?

astral303:
@theckman this blog post documents how to use chef-vault more securely, such that attributes are not leaked on node save:

http://jtimberman.housepub.org/blog/2014/12/24/quicktip-delete-attributes/

martinisoft:
I think this would be solved by moving the agent config into a Chef Resource. I'm planning to put up a PR for one in the near-ish future because I too don't like that it saves API keys in the clear.

olivielpeau:
@theckman @martinisoft: Could you have a look at the approach of #395 (see discussion in #345)?
It makes the cookbook pull the `api_key` and the `application_key` from the `run_state` first, so your wrapper cookbook could set the keys in the `run_state` and wouldn't leak the credentials.

PR #395, opened 2017-01-19
Defer evaluation of api and app keys + read from `run_state`

olivielpeau:
Based on #345 (thanks @mfischer), rebased on top of master, and minus the changes on the `compile_time` property in `dd-handler`. Details in the commit messages and in #345.

Tested on both Chef 11 and 12, seems to work fine.

Related to #377

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,9 @@ Datadog Cookbook
 
 Chef recipes to deploy Datadog's components and configuration automatically.
 
+**NB: This README may refer to features that are not released yet. Please check the README of the
+git tag/the gem version you're using for your version's documentation**
+
 Requirements
 ============
 - chef >= 10.14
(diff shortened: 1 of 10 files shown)


olivielpeau:
Hmm actually I realize now that enabling the datadog handler at compile time could defeat the purpose of pulling the keys from the `run_state`. Unless other recipes set the `run_state` at compile time before `dd-handler` is compiled the `run_state` won't be populated with the keys.

We could add an attribute to only enable the handler at converge time.

libraries/recipe_helpers.rb
```diff
@@ -0,0 +1,24 @@
...
+        run_state_or_attribute(node, 'api_key')
+      end
+
+      def application_key(node)
+        run_state_or_attribute(node, 'application_key')
+      end
+
+      private
+
+      def run_state_or_attribute(node, attr)
```
degemer: Should we rename `attr`? It's already used by Ruby (not in this context though) - http://ruby-doc.org/core-2.0.0/Module.html#method-i-attr

README.md
```diff
@@ -86,10 +86,16 @@ Usage
 
 1. Add this cookbook to your Chef Server, either by installing with knife or by adding it to your Berksfile:
   ```
-  cookbook 'datadog', '~> 2.1.0'
+  cookbook 'datadog', '~> 2.7.0'
   ```
-2. Add your API Key as a node attribute via an `environment` or `role` or by declaring it in another cookbook at a higher precedence level.
-3. Create an 'application key' for `chef_handler` [here](https://app.datadoghq.com/account/settings#api), and add it as a node attribute, as in Step #2.
+2. Add your API Key either:
```
degemer: Should we add at the top of the README that these instructions are not for 2.7.0?

recipes/dd-agent.rb
```diff
@@ -51,27 +51,25 @@
 # To add integration-specific configurations, add 'datadog::config_name' to
 # the node's run_list and set the relevant attributes
 #
-raise "Add a ['datadog']['api_key'] attribute to configure this node's Datadog Agent." if node['datadog'] && node['datadog']['api_key'].nil?
```
degemer: Should we keep it? We'd have to change the message. "Add a ['datadog']['api_key'] attribute or set the ['datadog']['api_key'] run_state to configure this node's Datadog Agent."

templates/default/datadog.conf.erb
```diff
@@ -141,11 +141,13 @@ if node['datadog']['legacy_integrations']
 end
 -%>
 
-<% if ! @extra_config.empty? -%>
+<% if ! node['datadog']['extra_config'].empty? -%>
 # ========================================================================== #
 # Other config options
 # ========================================================================== #
-  <% @extra_config.each do |k, v| -%>
+  <% @node['datadog']['extra_config'].each do |k, v| -%>
```
degemer: `@`?

degemer: Weird it's wasn't caught by the tests...

recipes/dd-agent.rb
olivielpeau: To keep it we'd need to evaluate it at execution time. I'll try that.

templates/default/datadog.conf.erb
olivielpeau: ha, good catch, apparently `@node` also works... Changing this to `node` anyway

README.md
olivielpeau: Done

recipes/dd-agent.rb
```diff
@@ -17,6 +17,13 @@
 # limitations under the License.
 #
 
+# Fail here at converge time if no api_key is set
+ruby_block 'datadog-api-key-set' do
+  block do
+    raise "Set ['datadog']['api_key'] as an attribute or on the node's run_state to configure this node's Datadog Agent." if Chef::Datadog.api_key(node).nil?
```
degemer: 🍰

PR #400, opened 2017-02-03
[dd-agent][windows] Fix agent version pinning

olivielpeau:
When the `agent_version` attribute is a string. Bug introduced in #368

Also added the test that would've caught the bug.

I'll release a bugfix version once this is merged.

### DIFF
--- a/recipes/_install-windows.rb
+++ b/recipes/_install-windows.rb
@@ -21,7 +21,7 @@ dd_agent_version =
   if node['datadog']['agent_version'].respond_to?(:each_pair)
     node['datadog']['agent_version']['windows']
   else
-    node['datadog']['agent_verison']
+    node['datadog']['agent_version']
   end
 
 # If no version is specified, select the latest package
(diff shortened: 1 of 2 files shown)


mlcooper:
Hi @olivielpeau I'm not sure I see what the fix was that you implemented, or what the bug was.  Was there an error message produced?  I do see a new spec test, but I do not see any code fixes in the cookbook.  Perhaps I'm missing something?

olivielpeau:
Hi @mlcooper, the issue was the typo on the attribute key, fixed here: https://github.com/DataDog/chef-datadog/blob/v2.8.1/recipes/_install-windows.rb#L24

mlcooper:
I see it now, thanks.  Sorry about that typo, thanks for catching it.

olivielpeau:
No problem! I think what really bit us here is that there had never been a test on the version pinning on Windows

mlcooper:
Did this test on the 2.7.0 tag not catch it?

https://github.com/DataDog/chef-datadog/blob/v2.7.0/spec/dd-agent_spec.rb#L266-L296

olivielpeau:
No unfortunately this test only checks when the `agent_version` is provided as a hash, and there was no test checking the case when the `agent_version` is provided as a string on Windows (there was one on Linux).

Issue #401, opened 2017-02-10
New datadog.conf settings for Traces


mlcooper:
Allow configuration of settings for Traces in datadog.conf

https://app.datadoghq.com/trace/docs/tutorials/configuration

PR #402, opened 2017-02-10
#401 - Allow configuration of Traces settings in datadog.conf

mlcooper:


attributes/default.rb
```diff
@@ -252,3 +252,16 @@
 default['datadog']['legacy_integrations']['nagios']['enabled'] = false
 default['datadog']['legacy_integrations']['nagios']['description'] = 'Nagios integration'
 default['datadog']['legacy_integrations']['nagios']['config']['nagios_log'] = '/var/log/nagios3/nagios.log'
+
+# Trace functionality settings
+default['datadog']['enable_trace_agent'] = false
+default['datadog']['extra_sample_rate'] = 1
+default['datadog']['max_traces_per_second'] = 10
+default['datadog']['receiver_port'] = 7777
```
talwai: we're actually deprecating 7777 for 8126 to be contiguous with statsd (though both work for the time being) . mind updating the default?

talwai:
LGTM, aside from the comment - would be nice if @olivielpeau could take a glance as well

attributes/default.rb
```diff
@@ -252,3 +252,16 @@
...
+
+# Trace functionality settings
+default['datadog']['enable_trace_agent'] = false
+default['datadog']['extra_sample_rate'] = 1
+default['datadog']['max_traces_per_second'] = 10
+default['datadog']['receiver_port'] = 8126
+default['datadog']['connection_limit'] = 2000
+
+# ddtrace python version
+default['datadog']['ddtrace_python_ver'] = nil
```
olivielpeau: could you use a full `ddtrace_python_version` attribute name here? (just trying to be consistent with all the other occurences of `*_version` in the attributes names of the cookbook)

same thing for `ddtrace_gem_version` below

PR #404, opened 2017-02-13
Fixed #403 - Guard new GPG key from being downloaded always

### LINKED ISSUE #403: remote_file[DATADOG_RPM_KEY_E09422B3.public] needs a guard if the key is already cached or already imported
https://github.com/DataDog/chef-datadog/blob/master/recipes/repository.rb#L56-L59 needs a guard. This resource is executed on every chef run even if the key has already been cached. The problem is not only downloading unnecessarily on every chef run, but the real issue comes into play when the remote source URL becomes unavailable. For example, if a node's remote RPM repo goes offline and the key can't be downloaded, it will cause the entire chef run to fail. Since the key is cached on the node already, this instance should not fail a chef run. Doubly so if the rpm gpg key has already been imported.

Chef run failure:
```
13T02:33:27+00:00] ERROR: remote_file[DATADOG_RPM_KEY_E09422B3.public] (datadog::repository line 56) had an error: Errno::ECONNRESET: Error connecting to https://******.com/nexus/content/repositories/Chef-Master/com/datadog/datadog_rpm_key.public/E09422B3/datadog_rpm_key.public-E09422B3.public - Connection reset by peer
```

Above key is already cached:
```
# ls -l /var/chef/cache/DATADOG_RPM_KEY_E09422B3.public 
-rw-r--r-- 1 root root 1662 Nov 23 18:12 /var/chef/cache/DATADOG_RPM_KEY_E09422B3.public
```

iancward:
- Moved chefspec examples for repo out of dd-agent_spec and into repository_spec
- Collapsed rhellions contexts since repository recipe doesn't care about python version
- Add a few missing examples

### DIFF
--- a/spec/repository_spec.rb
+++ b/spec/repository_spec.rb
@@ -67,7 +67,7 @@ describe 'datadog::repository' do
     end
 
     context 'version 5' do
-     cached(:chef_run) do
+      cached(:chef_run) do
         ChefSpec::SoloRunner.new(
           platform: 'centos', version: '5.10'
         ).converge(described_recipe)
--- a/spec/repository_spec.rb
+++ b/spec/repository_spec.rb
@@ -1,37 +1,66 @@
 describe 'datadog::repository' do
-  context 'rhellions' do
-    describe 'on versions 5.x and lower' do
-      cached(:chef_run) do
-        ChefSpec::SoloRunner.new(
-          :platform => 'centos',
-          :version => '5.8'
-        ) do |node|
-          node.set['languages'] = { 'python' => { 'version' => '2.4.3' } }
-        end.converge described_recipe
-      end
-
-      it 'sets up a yum repo' do
-        expect(chef_run).to create_yum_repository('datadog').with(
-          gpgkey: 'http://yum.datadoghq.com/DATADOG_RPM_KEY.public'
-        )
-      end
+  context 'on debianoids' do
+    cached(:chef_run) do
+      ChefSpec::SoloRunner.new(
+        platform: 'debian', version: '8.5'
+      ).converge(described_recipe)
     end
 
-    describe 'on versions 6.x and higher' do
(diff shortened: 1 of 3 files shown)


spec/repository_spec.rb
```diff
@@ -1,37 +1,66 @@
...
-        ) do |node|
-          node.set['languages'] = { 'python' => { 'version' => '2.4.3' } }
-        end.converge described_recipe
-      end
-
-      it 'sets up a yum repo' do
-        expect(chef_run).to create_yum_repository('datadog').with(
-          gpgkey: 'http://yum.datadoghq.com/DATADOG_RPM_KEY.public'
-        )
-      end
```
olivielpeau: any reason for removing this test on RHEL <= 5?

iancward: Hi @olivielpeau I can add it back, but the repository recipe doesn't have any logic dependent on platform version. The separate test for RHEL5 wasn't actually exercising anything specific, so I thought I'd condense the spec by removing it.

olivielpeau: Actually there's one RHEL5-specific thing: the yum repository has to be configured with a gpgkey downloaded through plain text `http` (the logic that handles that is in the default attributes file: https://github.com/DataDog/chef-datadog/blob/v2.8.1/attributes/default.rb#L85-L91).

iancward: Oooh.. good catch. This should be fixed in bc60110.

iancward:
It looks like https://travis-ci.org/DataDog/chef-datadog/builds/202371935 is (hopefully) caused by a temporary issue with Chef's Supermarket or connectivity to it.

olivielpeau:
Looks good now, thanks @iancward!

The CI failure was indeed a temp issue with Supermarket, just re-ran it and it passed.

PR #405, opened 2017-02-17
Added apm_enabled

jjasghar:
This will enable apm_enabled to be part of this cookbook.

Signed-off-by: JJ Asghar <jj@chef.io>

### DIFF
--- a/templates/default/datadog.conf.erb
+++ b/templates/default/datadog.conf.erb
@@ -37,6 +37,10 @@ proxy_user: <%= node['datadog']['web_proxy']['user'] %>
 proxy_password: <%= node['datadog']['web_proxy']['password'] %>
 <% end -%>
 
+<% if node['datadog']['apm_enable'] == true %>
+  apm_enable: true
+<% end -%>
+
 <% if node['datadog']['non_local_traffic'] == true %>
 non_local_traffic: true
 <% end -%>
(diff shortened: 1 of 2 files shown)


templates/default/datadog.conf.erb
```diff
@@ -37,6 +37,10 @@ proxy_user: <%= node['datadog']['web_proxy']['user'] %>
 proxy_password: <%= node['datadog']['web_proxy']['password'] %>
 <% end -%>
 
+<% if node['datadog']['apm_enable'] == true %>
+  apm_enable: true
```
robbkidd: This should be `apm_enabled` with a `d`. ( [code](https://github.com/DataDog/datadog-trace-agent/blob/dc580a0cd378bf0723e5af099605cf9d410a921f/config/agent.go#L196), [docs](https://app.datadoghq.com/trace/install) )

jjasghar: ❤️

robbkidd:
👍  Without this, I am faced with declaring `apm_enabled: true` as either a `legacy_integrations` or an `extra_config`

mlcooper:
Looks like a this will be taken care of with #401 which has some additional granular settings for APM and recipes for the require python/ruby packages

jjasghar:
Sounds good, thanks!

PR #409, opened 2017-03-01
Added Kubernetes recipe

xt99:
Having this recipe in place one can easily monitor Kubernetes with DataDog using "the chef way"


test/integration/datadog_kubernetes/serverspec/kubernetes_spec.rb
```diff
@@ -0,0 +1,33 @@
+# Encoding: utf-8
```
olivielpeau: instead of an integration test here, could you write a spec test? (you can take https://github.com/DataDog/chef-datadog/blob/v2.8.1/spec/integrations/docker_daemon_spec.rb as an example)

Spec tests are way lighter to run and can cover the integration recipes very well

xt99:
Done:

```
$ rspec spec/integrations/kubernetes_spec.rb
..........

Finished in 2.66 seconds (files took 3.78 seconds to load)
10 examples, 0 failures

ChefSpec Coverage report generated...

  Total Resources:   7
  Touched Resources: 7
  Touch Coverage:    100.0%
```

.kitchen.yml
```diff
@@ -394,6 +394,19 @@ suites:
               my_consumer:
                 my_topic: [0, 1, 4, 12]
 
+- name: datadog_kubernetes
```
olivielpeau: actually could you remove the integration test and the related dir `test/integration/datadog_kubernetes`?

These tests are deprecated for integration recipes, we're just keeping the tests there until we move them all to spec tests.

Once you've removed them I'll merge, thanks 👍

PR #410, opened 2017-03-07
[windows] Add option to support EXE Agent installer

olivielpeau:
The EXE installer handles per-user to per-machine upgrades
gracefully, including edges cases that the present cookbook can't
reasonably handle on its own.

For this reason, we add the option to use the exe installer when needed
(i.e. for upgrades from Agent <= 5.10.1 to Agent >= 5.12.0).

Also added related documentation

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -111,6 +111,13 @@ default['datadog']['yumrepo_gpgkey_new'] = "#{yum_protocol}://yum.datadoghq.com/
 # Expected checksum to validate correct agent installer is downloaded (Windows only)
 default['datadog']['windows_agent_checksum'] = nil
 
+# Set to `true` to use the EXE installer on Windows, recommended to gracefully handle upgrades from per-user
+# to per-machine installs for all known use cases. We recommend setting this option to `true` for Agent upgrades from
+# versions <= 5.10.1 to versions >= 5.12.0.
+# The EXE installer exists since Agent release 5.12.0
+# If you're already using version >= 5.12.0 of the Agent, leave this to false.
+default['datadog']['windows_agent_use_exe'] = false
+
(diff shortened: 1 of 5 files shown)


attributes/default.rb
```diff
@@ -111,6 +111,13 @@
 # Expected checksum to validate correct agent installer is downloaded (Windows only)
 default['datadog']['windows_agent_checksum'] = nil
 
+# Set to `true` to use the EXE installer on Windows, recommended to gracefully handle upgrades from per-user
+# to per-machine installs for all known use cases. We recommend setting this option to `true` for Agent upgrades from
```
derekwbrown: I'm not sure I'd say "all known use cases".  That's pretty broad :)

recipes/_install-windows.rb
```diff
@@ -25,28 +25,48 @@
   end
 
 # If no version is specified, select the latest package
-dd_agent_msi = dd_agent_version ? "ddagent-cli-#{dd_agent_version}.msi" : 'ddagent-cli.msi'
-temp_file = ::File.join(Chef::Config[:file_cache_path], 'ddagent-cli.msi')
+dd_agent_installer_basename = dd_agent_version ? "ddagent-cli-#{dd_agent_version}" : 'ddagent-cli'
```
olivielpeau: at this point the default installer URL is not different for the MSI (and the EXE has the same URL, except that it has an `.exe` extension).

In general though if we're concerned about the user experience of automatic upgrades failing when the default latest package is downloaded (shouldn't happen with this cookbook but could happen with user-maintained scripts) then we may consider using a different URL for the new installer.

olivielpeau:
@degemer I've added a few commits since your review, could you have one last look at the changes? :)

Issue #194, opened 2015-04-07
Synchronize with current dd-agent configuration examples (Numero Dos)


miketheman:
Following #154, continue with sync for another milestone.
- [ ] hdfs
- [ ] iis
- [ ] jenkins
- [ ] kafka_consumer
- [ ] kyototycoon
- [ ] lighttpd
- [ ] memcache
- [ ] mongo
- [ ] mysql
- [ ] network
- [ ] nginx
- [x] postgres (also see #140)
- [ ] rabbitmq
- [ ] redisdb
- [ ] riak
- [ ] tomcat
- [x] varnish
- [ ] zookeeper

mcg:
Any chance this is still being worked on? Trying to use the Nginx module and it's out of date, doesn't support "ssl_validation".

iancward:
@mcg in case you didn't see #194, the `ssl_validation` option is now supported for the nginx integration.

PR #417, opened 2017-03-27
Address #194 for nginx

iancward:
Update nginx integration template with features expressed in current integration.

### DIFF
--- a/templates/default/nginx.yaml.erb
+++ b/templates/default/nginx.yaml.erb
@@ -1,6 +1,12 @@
 instances:
 <% @instances.each do |i| -%>
   - nginx_status_url: <%= i['nginx_status_url'] %>
+    <%= "ssl_validation: #{i['ssl_validation']}" unless i['ssl_validation'].nil? -%>
+  <%# don't print user/password if either is nil -%>
+  <% unless i['user'].nil? || i['password'].nil? -%>
+    user: <%= i['user'] %>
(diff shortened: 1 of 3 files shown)


templates/default/nginx.yaml.erb
```diff
@@ -1,6 +1,12 @@
 instances:
 <% @instances.each do |i| -%>
   - nginx_status_url: <%= i['nginx_status_url'] %>
+    <%= "ssl_validation: #{i['ssl_validation']}" unless i['ssl_validation'].nil? -%>
```
olivielpeau: IMO the indentation and format of this line could break easily if we make changes to it in the future, could you rather write:
```erb
<% unless i['ssl_validation'].nil? -%>
    ssl_validation: i['ssl_validation']
<% end -%>
```

What do you think?

iancward: I waffled on whether to do it that way, but now that I think about it, what you put above will also be more consistent with the rest of the template. Let me take care of that.

iancward: Oh! I had not even thought of chefspec; let me add that as well.

iancward: @olivielpeau I've pushed a5e1f25 with the recommended changes; I also added spec tests and updated the examples in the nginx recipe to demonstrate the new features.

Issue #418, opened 2017-03-28
enable_trace_agent Attribute Breaks Configuration for Service Discovery


bflad:
Hello there!

We tried switching our attributes from:
```ruby
default['datadog']['extra_config'] = {
    'apm_enabled' => 'true',
    'service_discovery_backend' => 'docker',
    'sd_config_backend' => 'consul',
    'sd_backend_host' => '127.0.0.1',
    'sd_backend_port' => '8500'
}
```

To the new `enable_trace_agent` attribute via:

```ruby
default['datadog']['enable_trace_agent'] = true 
default['datadog']['extra_config'] = {
    'service_discovery_backend' => 'docker',
    'sd_config_backend' => 'consul',
    'sd_backend_host' => '127.0.0.1',
    'sd_backend_port' => '8500'
}
```

However in the configuration template for `enable_trace_agent` (https://github.com/DataDog/chef-datadog/blob/master/templates/default/datadog.conf.erb#L64-L75), it inserts new subsection headers into the configuration file, which breaks service discovery by not switching the configuration subsection back to `[Main]`:

```python
# Generated by Chef, local modifications will be overwritten

[Main]
...
## Trace settings
apm_enabled: true

[trace.sampler]
extra_sample_rate: 1
max_traces_per_second: 10

[trace.receiver]
receiver_port: 8126
connection_limit: 2000
...
# ========================================================================== #
# Other config options
# ========================================================================== #
service_discovery_backend: docker
sd_config_backend: consul
sd_backend_host: 127.0.0.1
sd_backend_port: 8500
```

The most simple fix would be to add a `[Main]` line at the end of the if statement in the template. However it would be also nice if service discovery had proper attributes that would be placed above the trace agent config.

olivielpeau:
I'll go ahead and close this since you fixed the issue on the trace agent config option @bflad, thanks again!

The changes on the service discovery attributes will be released with the next minor

PR #419, opened 2017-03-28
Add [Main] section header to end of datadog.conf enable_trace_agent template conditional

bflad:
Simplest fix for #418.

### DIFF
--- a/templates/default/datadog.conf.erb
+++ b/templates/default/datadog.conf.erb
@@ -72,6 +72,8 @@ max_traces_per_second: <%= node['datadog']['max_traces_per_second'] %>
 [trace.receiver]
 receiver_port: <%= node['datadog']['receiver_port'] %>
 connection_limit: <%= node['datadog']['connection_limit'] %>
+
+[Main]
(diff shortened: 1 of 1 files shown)


degemer:
Thanks for reporting and submitting a fix @bflad ! We'll review it shortly (today or tomorrow).

templates/default/datadog.conf.erb
```diff
@@ -72,6 +72,8 @@ max_traces_per_second: <%= node['datadog']['max_traces_per_second'] %>
 [trace.receiver]
 receiver_port: <%= node['datadog']['receiver_port'] %>
 connection_limit: <%= node['datadog']['connection_limit'] %>
+
+[Main]
```
olivielpeau: Actually this works, but I think it would be more human-readable to have the whole `Main` section in one block.

Could you move the whole trace-agent-related part of the config at the end of the file so that the `Main` section is not split?

bflad: I've pushed up that change!

PR #420, opened 2017-03-28
Add service discovery attributes and datadog.conf template to enable

bflad:
Semi-related to #418, this adds the service discovery attributes as well as keeps them above the trace agent configuration which breaks the `[Main]` datadog.conf configuration section.


olivielpeau:
Could you fix the conflicts @bflad? :)

bflad:
Updated from the release changes yesterday. 😄

olivielpeau:
Thanks! Merging

PR #424, opened 2017-04-07
Mongo & Elasticsearch: extend config templates with more options

gkze:
We need these options in production and are currently overriding the config. We'd like to have these available to configure.

### DIFF
--- a/templates/default/elastic.yaml.erb
+++ b/templates/default/elastic.yaml.erb
@@ -8,7 +8,7 @@ instances:
     pshard_stats: <%= i["pshard_stats"] %>
     <% end %>
     <% unless i["cluster_stats"].nil? %>
-      cluster_stats: <%= i["cluster_stats"] %>
+    cluster_stats: <%= i["cluster_stats"] %>
     <% end %>
     <% unless i["is_external"].nil? %>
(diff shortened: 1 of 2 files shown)


templates/default/elastic.yaml.erb
```diff
@@ -7,6 +7,9 @@ instances:
     <% unless i["pshard_stats"].nil? %>
     pshard_stats: <%= i["pshard_stats"] %>
     <% end %>
+    <% unless i["cluster_stats"].nil? %>
+      cluster_stats: <%= i["cluster_stats"] %>
```
olivielpeau: the indentation is not correct here: `cluster_stats` needs to have the same indentation as the other parameters. Could you correct it?

(same comment for the `ssl`  options of `mongo.yaml.erb`)

gkze:
Good catch @olivielpeau, thank you. Corrected.

olivielpeau:
For some reason I can't make the CI to re-run against a merge of this PR into the _current_ state of master, so the build still fails.

Merging this, I'll check that the resulting CI build on master is successful.

Thanks @gkze!

gkze:
Thank you @olivielpeau 👍

PR #427, opened 2017-04-25
make package names compatible

gmmeyer:
many integrations have an underscore in their name. They should install with a dash, underscores are forbidden in most operating systems and omnibus builds it with the dash.

### DIFF
--- a/recipes/integrations.rb
+++ b/recipes/integrations.rb
@@ -30,7 +30,9 @@ include_recipe 'datadog::repository' if node['datadog']['installrepo']
 #   'url' => 'http://localhost:22222'
 # }
 node['datadog']['extra_packages'].each do |name, options|
-  package options['name'] do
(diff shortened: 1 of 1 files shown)


recipes/integrations.rb
```diff
@@ -30,7 +30,7 @@
 #   'url' => 'http://localhost:22222'
 # }
 node['datadog']['extra_packages'].each do |name, options|
-  package options['name'] do
+  package options['name'].sub('_', '-') do
```
olivielpeau: couldn't we cover this case with better documentation here and in the `attributes/default.rb` file?

Making the substitution ourselves in this recipe forces us to support this substitution "forever" (changing the behavior later would be a breaking change)

gmmeyer: We should and I'm just going to close this because of that

PR #430, opened 2017-04-28
Fixed #429 - Add Recipe for disk integration

### LINKED ISSUE #429: [disk] configuration for disk integration?
While the disk integration for the Datadog agent doesn't require any configuration by default, it looks like you are able to set up a file to override the defaults (e.g. to exclude certain file systems or mounts from metrics collection). It doesn't look like this cookbook provides the ability to add this custom configuration.
Did I miss something or does this cookbook not have that feature?

iancward:


spec/integrations/disk_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        disk: {
+          instances: [
+            {
+              use_mount: 'yes',
+              excluded_filesystems: ['tmpfs'],
+              excluded_disks: ['/dev/sda1', '/dev/sda2'],
+              excluded_disk_re: '/dev/sde.*',
+              tag_by_filesystem: 'no',
+              excluded_mountpoint_re: '/mnt/somebody-elses-problem.*',
+              all_partitions: 'no'
```
olivielpeau: Even though the disk check appears to correctly handle strings for boolean parameters, it's recommended to set them as actual booleans. Could you use booleans for `use_mount`, `tag_by_filesystem` and `all_partitions`? (and update the example you provided in the recipe accordingly).

iancward: Hi @olivielpeau: I actually took the examples directly from http://docs.datadoghq.com/integrations/disk/, but I've made the requested changes in bdec562.

olivielpeau: Thanks!

Actually the examples in the docs are correct, in yaml `yes` and `no` are booleans (but `'yes'` and `'no'` are strings).

olivielpeau:
@iancward One last thing: could you fix the small rubocop failure? https://travis-ci.org/DataDog/chef-datadog/builds/228122567#L266

spec/integrations/disk_spec.rb
iancward: Ah, right. I think I had quoted those because the 'yes' and 'no' string attributes were getting converted to true/false booleans in the file and the spec was failing. Using true/false directly definitely removes confusion here.

iancward:
Oopsie. Fixed in f3b925d .
I had originally tried using a `do`/`end` block instead of the `{` and `}` (which is supposed to be used for single line things), but it couldn't get it to work properly. I had the same experience with the chefspec in the for the nginx integration.

olivielpeau:
Thanks @iancward! PR looks good, merging.

The issue you had with `do`/`end` is probably related to the lower precedence of `do`/`end` compared to `{`/`}`... I'd say it's fine to use curly braces here anyway.

PR #433, opened 2017-05-04
[trace-agent] Leave the default enabled/disabled behavior to the agent

olivielpeau:
Since agent 5.13.0, the trace agent is enabled by default.

This cookbook should allow explicitly enabling or disabling the trace
agent, however, when the attribute that controls this is not set, the
cookbook should leave it to the agent to decide whether to run the
trace agent.

This change modifies the default value of an attribute. Technically
this is a breaking change that requires a major version bump of the
cookbook, however I think this change actually improves backward
compatibility and can go with a minor version of the cookbook.

Follow-up to https://github.com/DataDog/chef-datadog/pull/428, cc @ed-flanagan

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -272,7 +272,11 @@ default['datadog']['sd_template_dir'] = '/datadog/check_configs'
 default['datadog']['service_discovery_backend'] = nil
 
 # Trace functionality settings
-default['datadog']['enable_trace_agent'] = false
+# Set `enable_trace_agent` to:
+# * `true` to explicitly enable the trace agent and customize its settings
+# * `false` to explicitly disable it
+# Leave it to `nil` to let the agent's default behavior decide whether to run the trace-agent
+default['datadog']['enable_trace_agent'] = nil
 default['datadog']['extra_sample_rate'] = 1
 default['datadog']['max_traces_per_second'] = 10
 default['datadog']['receiver_port'] = 8126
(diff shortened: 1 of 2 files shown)


ed-flanagan:
Makes sense.
However, when `nil` (i.e. default), the trace.sampler/trace.receiver templates will not be used. Would the situation of setting those, but not `enable_trace_agent` make sense?
Perhaps restricting the if/else block to just the `apm_enabled` directive would be more "flexible", e.g.
```erb
<% if node['datadog']['enable_trace_agent'].is_a?(TrueClass) || node['datadog']['enable_trace_agent'].is_a?(FalseClass) -%>
apm_enabled: <%= node['datadog']['enable_trace_agent'] ? 'true' : 'false' %>
<% end -%>

olivielpeau:
@ed-flanagan Thanks for your feedback! The one thing I disliked about setting the `trace-agent` settings even when it wasn't explicitly enabled was that the cookbook would always set default values on the trace-agent settings.
IMO when the user doesn't define any settings the cookbook shouldn't enforce its own defaults, instead it should let the trace-agent choose its own defaults.

I've pushed a commit that changes that behavior and implements what you suggested, let me know what you think

ed-flanagan:
Totally agree. If `trace-agent` already has its own defaults I feel it's better to not redefine them in the cookbook. It's less opinionated and removes the need to update the defaults if/when `trace-agent`'s own defaults change.
I like the new changes! 👍

olivielpeau:
Thanks @ed-flanagan, merging this then!

PR #435, opened 2017-05-05
Allow to specify trace env

krasnoukhov:
Our APM just recently started reporting to "env:none". Adding `tags` to a `datadog.conf` does not help for whatever reason. I noticed that configuring trace like this is missing from chef, so here we go.

https://app.datadoghq.com/apm/docs/tutorials/environments

### DIFF
--- a/templates/default/datadog.conf.erb
+++ b/templates/default/datadog.conf.erb
@@ -183,6 +183,11 @@ end
 apm_enabled: <%= node['datadog']['enable_trace_agent'] %>
 <% end -%>
 
+[trace.config]
+<% unless node['datadog']['trace_env'].nil? -%>
+env: <%= node['datadog']['trace_env'] %>
+<% end -%>
+
 [trace.sampler]
 <% unless node['datadog']['extra_sample_rate'].nil? -%>
 extra_sample_rate: <%= node['datadog']['extra_sample_rate'] %>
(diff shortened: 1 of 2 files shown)


olivielpeau:
Hi @krasnoukhov, thanks for adding this! Actually I just merged #433 and it conflicts with your PR unfortunately, could you rebase yours? (and maybe change the template condition to `<% unless node['datadog']['trace_env'].nil? -%>` to be consistent with #433, unless you think your condition makes more sense)

krasnoukhov:
@olivielpeau Done

templates/default/datadog.conf.erb
```diff
@@ -183,6 +183,11 @@ end
 apm_enabled: <%= node['datadog']['enable_trace_agent'] %>
 <% end -%>
 
+[trace.config]
+<% unless node['datadog']['trace_env'].nil? -%>
+env = <%= node['datadog']['trace_env'] %>
```
olivielpeau: Could you use `env: ` instead of `env = ` here? I know the 2 syntaxes _should_ be similar in the `ini` format but still I'd rather be consistent with the other options

krasnoukhov: Oh yeah totally missed that. I copied the line from documentation initially, so that could be changed as well

olivielpeau:
Thanks! Merging

PR #436, opened 2017-05-12
Add recipe for SNMP

mlcooper:


olivielpeau:
Thanks @mlcooper, looks great! I've merged this in 0dd7f9f (authorship should be preserved), this will be released with the next minor

Issue #439, opened 2017-05-31
Windows cookbook version recommendation


mlcooper:
With the release of 2.10.1, is this statement in the README still valid, or will they work if you jump to depending on 3.x of the `windows` cookbook?
```
With Chef >= 12.6 and the windows cookbook >= 1.39.0, Agent upgrades are known to fail. For Chef>=12.6 users on Windows, we recommend pinning the windows cookbook to a lower version (~> 1.38.0 for instance).
```

olivielpeau:
Hi @mlcooper!

This statement is still valid unfortunately. On Windows, if you're using `Chef >= 12.6` and need to use `windows >= 1.39`, `3.0` included, you'll have to use the `remove-dd-agent` recipe before upgrading the agent.

The present cookbook tries to work around multiple edge cases caused by the very inconsistent behavior of the `windows_package` resource depending on the version of Chef (between 12.0 and 12.6) and the version of the `windows` cookbook.

I think the only way to solve this once and for all is to drop support for Chef<=12.6 on Windows. That'll allow using the core resources of Chef only (and drop the dependency on `windows`), and use a `windows_package` resource that just "works".

This will be part of the next major release of the cookbook.

mlcooper:
I think that's a great idea to ultimately drop the dependency on the `windows` cookbook.  

Since support for Ruby v2.1 was dropped on April 1, 2017, the earliest Chef client that runs a supported version of Ruby is now v12.14.60, so I think you can certainly make the case for dropping support for Chef <= 12.6.

PR #323, opened 2016-07-08
fixing notification handlers

someara:
Hello!

This PR will make chef-client go a lottapercent faster on hosts with multiple checks that need to be restarted.

-s

PR #450, opened 2017-07-10
Chef 12.7+ Monitor Resource

martinisoft:
This re-writes the current monitor resource to be a Chef 12.7+ based resource. Given that Chef 10 and 11 are deprecated and 12 will be deprecated by April of next year I think it's time this cookbook caught up to the newer patterns with Chef. It should be entirely backwards compatible with the current resource usage for consumers of this cookbook, however they will need to upgrade to _at least_ Chef 12.7 in order to use this cookbook.

If needed, I can help document the resource in the README as well. This cookbook passes basic linting checks, but I was unable to run the spec suite due to issues with installing the coveralls into the ChefDK. Speaking of, I would suggest someone familiar with testing the cookbook look at adding a TESTING.md document to the cookbook to document that.


martinisoft:
Seems the travis builds aren't matching rubocop locally so I am not sure how to proceed with resolving those build issues.

mlcooper:
This PR would incorporate nicely with taking care of https://github.com/DataDog/chef-datadog/issues/439

martinisoft:
I have rebased this to catch it back up with master and changed the version to Chef 12.7 because of the bugs with custom resources that got resolved by 12.7.

resources/monitor.rb
```diff
@@ -1,15 +1,64 @@
...
+      init_config: new_resource.init_config,
+      instances:   new_resource.instances,
+      version:     new_resource.version
+    )
+    cookbook new_resource.cookbook
+    sensitive true
+    notifies :restart, 'service[datadog-agent]', :delayed if node['datadog']['agent_start']
+  end
+
+  service 'datadog-agent' do
```
olivielpeau: Since this PR will require a major version bump of the cookbook, I think it's time to remove the `service` resource from the custom resource and make invocations of `datadog_monitor` notify `service[datadog-agent]` directly instead (see #467 and #323 for some background on the current related issue).

.kitchen.yml
```diff
@@ -15,8 +15,8 @@ platforms:
       debian-7.7
     )
     chef_versions = %w(
-      12.2.1
-      11.18.6
+      12.5.1
```
olivielpeau: this should be a `12.7` version?

martinisoft: Good catch. I'll address this.

resources/monitor.rb
martinisoft: Ah, that is fair 👍 

Will fix in another commit.

martinisoft:
I've done a rebase against current master to get fresh tests going and have addressed the feedback in the review. Can you take a look again @olivielpeau and I'll make any additional modifications required.

A future build on top of this if you are willing to go higher on Chef Client versions is to jump to 12.10 and use an accumulator to eliminate the possibility of resource cloning. I'm working on a blog post about how we achieved this with this very cookbook by building a wrapper resource.

olivielpeau:
Thanks @martinisoft! I'd be very interested in your blog post when it's ready :)

By the way the travis CI is still failing, looks like the default value of `nil` on the `version` property doesn't work for the type of the property you specified.

About the requirement of Chef >= 12.10, to be honest I don't have a very good sense of which versions of Chef the users of this cookbook run. Will depend on the benefits that this new requirement would bring. Since Chef 12 in general is not EOL'ed yet I'd like to keep requirements as low as possible though.

mlcooper:
If with this PR the version of chef client required is raised to 12.7, then unless the datadog cookbook uses any of these windows specific resources, https://github.com/chef-cookbooks/windows/tree/master/resources, the dependency on the `windows` cookbook should be able to be dropped.

martinisoft:
> By the way the travis CI is still failing, looks like the default value of nil on the version property doesn't work for the type of the property you specified.

I have corrected those @olivielpeau. A better re-factor there would be to use the new [property_is_set?](https://docs.chef.io/custom_resources.html#property-is-set) method in Chef 12 that will let you avoid nil checks and other quirks, but that is for another PR.

> About the requirement of Chef >= 12.10, to be honest I don't have a very good sense of which versions of Chef the users of this cookbook run. Will depend on the benefits that this new requirement would bring. Since Chef 12 in general is not EOL'ed yet I'd like to keep requirements as low as possible though.

True. Also as @mlcooper points out we can drop the windows cookbook requirement as well since the vast majority of the resources are now included in Chef as of 12.6 so I have taken the liberty of also cleaning this up in the PR.

irabinovitch:
@olivielpeau Thoughts on next steps for this one?

martinisoft:
I will have to rebase again to get it into a state where you can merge, but I'd like to see if this can get merged sooner rather than later. Chef 12 will become EOL in less than 5 months now.

martinisoft:
Merged in and resolved the new conflicts. I can do another re-factor beyond this one that will push you up higher to Chef 12.17+ which drops a lot more required code as well.

martinisoft:
Looks like I missed a couple things in the logs change merge which now look to be fixed. 👍

martinisoft:
Checking back in close to a year later, is there any interest in this PR from maintainers? If so I'll look at rebasing against master again. Chef 12 is EOL as of April so I'd suggest merging this PR and putting in deprecation notices and work toward a major version that only supports Chef 13+

martinisoft:
I have caught it up to master. Would suggest this as a breaking release since even Chef 12 is EOL. Lots and lots of code can be removed from this cookbook if you push to a Chef 13+ release.

martinisoft:
I've personally written an internal version of this cookbook with only two resources in about 200 lines of total code with zero recipes. I can port part of it to a Chef 13+ rewrite if you'd like.

martinisoft:
@olivielpeau Once this is merged I can do a new PR to introduce an agent resource to install and configure the 6.0 agent. It'll clean up even more code and dependencies and set the stage for Chef 14 compatibility.

olivielpeau:
> I would still highly suggest not supporting Chef 12 any time soon as in make 3.x your last Chef 12 supported release

Sounds good 👍  I'll mention this plan in the release notes of 3.0.

> Once this is merged I can do a new PR to introduce an agent resource to install and configure the 6.0 agent. It'll clean up even more code and dependencies and set the stage for Chef 14 compatibility.

Do you see this new resource as deprecating the `datadog::dd-agent` recipe, or at least that users would be encouraged to use the new resource directly instead of including the `datadog::dd-agent` recipe?

martinisoft:
> Do you see this new resource as deprecating the datadog::dd-agent recipe, or at least that users would be encouraged to use the new resource directly instead of including the datadog::dd-agent recipe?

Correct. Ideally, this cookbook should have _zero_ recipes and only resources by the time you get to Chef 14. A good first step will be to introduce the resource and mark the recipe for deprecation. The recipe will use the new resource initially, then be removed in the next major release. All of the integration recipes can go away including the vast majority of the attributes.

In the end, the way the cookbook is currently written, it is extremely difficult to maintain. Just check out the PRs only adding or tweaking a recipe here or there, it's aging fast and creating more code that can be simplified into one resource.

olivielpeau:
I agree that we should avoid all the integration-specific recipes that only drop a config file. Moving the cookbook towards more resources and fewer recipes sounds great to me as well 👍 

Working on getting 2.17 out, I'll then merge this PR for 3.0, happy to continue this discussion then on a new PR.

remeh:
Please note that the CHANGELOG for release 2.18.0 has been updated accordingly: https://github.com/DataDog/chef-datadog/pull/589

martinisoft:
I'm seeing some activity on this PR which is great though at this point it might be better to re-work all of this because in just over a month Chef 13 will be EOL per their yearly major release cycle. This puts Chef 12 which has been EOL since last year completely out of the picture and possibly Chef 13. It will GREATLY simplify your Windows based installs to use Chef 14 since a significant amount of work was done in Chef 13 and 14 to vastly improve those resources and support.

martinisoft:
Also supporting just Agent 6 will vastly simplify what would be needed to add/remove agent checks since it supports the `.d` folder pattern with exception of the datadog.conf file.

remeh:
Hey @martinisoft 
First, I have to thank you for your contributions, they're very helpful!
Indeed we're late with Chef versions... However, we still have customers using those old versions!
Basically, we plan to better support Chef 12 in chef-datadog 3.0 and then to quickly move to a better support of Chef 13 (and 14) with chef-datadog 4.0

metadata.rb
```diff
@@ -23,12 +23,15 @@
   supports os
 end
 
-depends          'apt' # We recommend '>= 2.1.0'. See CHANGELOG.md for details
-depends          'chef_handler', '>= 1.2' # We recommend '~> 1.3' with Chef < 12. See CHANGELOG.md for details
-depends          'windows' # We recommend '< 1.39.0' if running Chef >= 12.6. See README.md for details
-depends          'yum', '>= 3.0' # Use '~> 3.0' with Chef < 12
-
-suggests         'sudo' # ~FC052
+depends    'chef_handler'
```
olivielpeau: @remeh any reason to remove the version constraint on `chef_handler`?

metadata.rb
```diff
@@ -23,12 +23,15 @@
...
+# In order to support Chef < 12.9, we need apt_repository which is available
+# in apt < 6.0.0. This will be removed for chef-datadog 4.0 while dropping the support
+# of Chef 12.
+depends    'apt', '< 6.0.0'
```
olivielpeau: @remeh the dependency constraints we define in `metadata.rb` should be as loose as possible, since they're constraints that are applied on the whole Chef environment of a given user.
If we set the `apt` constraint only to support Chef < 12.9, we should not set the constraint here but instead document this additional constraint in the readme for Chef 12.9 users. For example, this will allow Chef 14 users to still use the `datadog` cookbook with apt `6.0`.
Does that make sense? Happy to discuss this offline as well

metadata.rb
```diff
@@ -23,12 +23,15 @@
...
+# In order to support Chef < 12.14, we need yum_repository which is available
+# in yum < 5.0. This will be removed for chef-datadog 4.0 while dropping the support
+# of Chef 12.
+depends    'yum', '< 5.0'
```
olivielpeau: @remeh same comment here

resources/monitor.rb
```diff
@@ -1,16 +1,56 @@
...
 # is evaluated.
-attribute :init_config, :kind_of => Hash, :required => false, :default => {}
-attribute :instances, :kind_of => Array, :required => false, :default => []
-attribute :logs, :kind_of => Array, :required => false, :default => []
-attribute :version, :kind_of => Integer, :required => false, :default => nil
-attribute :use_integration_template, :kind_of => [TrueClass, FalseClass], :required => false, :default => false
+property :init_config, [Hash, nil], required: false, default: {}
+property :instances, Array, required: false, default: []
+property :version, [Integer, nil], required: false, default: nil
+property :use_integration_template, [TrueClass, FalseClass], required: false, default: false
```
olivielpeau: @remeh small note, feel free to address in future PR: it would make sense to make the default `true` here, goes in the direction of promoting the generic `datadog_monitor` resource with standard template over the integration-specific recipes

resources/monitor.rb
```diff
@@ -1,16 +1,56 @@
...
+action :add do
+  log "Adding monitoring for #{new_resource.name}" do
+    level :debug
+  end
+
+  template ::File.join(node['datadog']['config_dir'], 'conf.d', "#{new_resource.name}.yaml") do
+    if node['platform_family'] == 'windows'
+      owner 'Administrators'
+      rights :full_control, 'Administrators'
+      inherits false
```
olivielpeau: @remeh this will have to be updated in light of your changes here: https://github.com/DataDog/chef-datadog/pull/588

resources/monitor.rb
```diff
@@ -1,16 +1,56 @@
...
+property :version, [Integer, nil], required: false, default: nil
+property :use_integration_template, [TrueClass, FalseClass], required: false, default: false
+property :logs, [Array, nil], required: false, default: []
+
+action :add do
+  log "Adding monitoring for #{new_resource.name}" do
+    level :debug
+  end
+
+  template ::File.join(node['datadog']['config_dir'], 'conf.d', "#{new_resource.name}.yaml") do
```
olivielpeau: @remeh unfortunately we still need the distinction between `config_dir` and `agent6_config_dir` that the current implementation has

resources/monitor.rb
```diff
@@ -1,16 +1,56 @@
...
+      version:     new_resource.version,
+      logs:        new_resource.logs
+    )
+    cookbook new_resource.cookbook
+    sensitive true
+  end
+end
+
+action :remove do
+  confd_dir = ::File.join(node['datadog']['config_dir'], 'conf.d')
```
olivielpeau: @remeh ditto

README.md
```diff
@@ -79,8 +67,6 @@ Installs the Datadog agent on the target system, sets the API key, and start the
 **Notes for Windows**:
 
 * With Chef >= 12.6 _and_ the `windows` cookbook >= 1.39.0, Agent upgrades are known to fail.
-  For Chef>=12.6 users on Windows, we recommend pinning the `windows` cookbook to a lower version (`~> 1.38.0` for instance).
-
   If that's not an option, a known workaround is to use the `remove-dd-agent` recipe (since the `2.5.0` version of the present cookbook) to uninstall the Agent
```
olivielpeau: @remeh this sentence doesn't make sense anymore since you removed the previous one

martinisoft:
> Indeed we're late with Chef versions... However, we still have customers using those old versions!
Basically, we plan to better support Chef 12 in chef-datadog 3.0 and then to quickly move to a better support of Chef 13 (and 14) with chef-datadog 4.0

Definitely a super long tail of support there. Keep in mind Chef 12 ships with an EOL Ruby as well so for the security conscious that would be a bigger deal given the recent chunks of CVEs that went out for Ruby and Rubygems. This PR can be fixed up for merge, but as a bit of advice, consider making a major version very soon which draws the line to _at least_ Chef 13 which will let you tighten up the dependencies and cleanup the resources even further. If you folks want I can show you our internal version of this which has an agent install, configuration, and check resource and no recipes to get things up and running super quickly in a Chef 13 or 14 environment. So much simpler to manage and covers the vast majority of your use cases.

metadata.rb
remeh: Removing the version constraint bump to use the latest one `3.0.3` which fills the need of one card of our todo-list about having a better usage of the chef-handler dep.
To be more precautious I could add the constraint `~> 3.0` if we really want to be sure that we stay with 3.x but as chef-handler is now part of Chef, not sure it's needed.
Please lmk if it's a misunderstanding.

resources/monitor.rb
remeh: restored

README.md
```diff
@@ -375,6 +363,24 @@ We are not making use of data_bags in this recipe at this time, as it is unlikel
 
 For more deployment details, visit the [Datadog Documentation site](http://docs.datadoghq.com/).
 
+### Chef 12
+
+Depending of the Chef 12 version you're using, you will have to add some extra
+dependency constraint.
```
olivielpeau: `contraints`

remeh:
Hey @martinisoft , again thanks a lot for your involvement! I'll finally be able to merge this. Later, we'll also work on a next major version dropping the support of Chef 12, if you have hints on further breaking changes to do so, we're definitely open to them. Thanks again 🙇!

PR #453, opened 2017-07-13
Avoid a resource cloning warning

martinisoft:
This is already done in the apt default recipe which is called above and it has been there for some time now. Resource cloning is currently a warning and in Chef 13.2 it has changed behavior which causes unpredictable resource execution now if you clone resources instead of editing them with the new find_resource and edit_resource helpers.

### DIFF
--- a/recipes/repository.rb
+++ b/recipes/repository.rb
@@ -21,11 +21,6 @@ case node['platform_family']
 when 'debian'
   include_recipe 'apt'
 
-  package 'install-apt-transport-https' do
-    package_name 'apt-transport-https'
-    action :install
-  end
-
   # Trust new APT key
   execute 'apt-key import key 382E94DE' do
     command 'apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 A2923DFF56EDA6E76E55E492D3A80E30382E94DE'
--- a/spec/repository_spec.rb
+++ b/spec/repository_spec.rb
@@ -10,10 +10,6 @@ describe 'datadog::repository' do
       expect(chef_run).to include_recipe('apt::default')
     end
 
-    it 'installs apt-transport-https' do
-      expect(chef_run).to install_package('install-apt-transport-https')
-    end
-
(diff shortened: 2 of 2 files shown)


martinisoft:
This was originally suggested in #388 but I think removal is a better option than just side-stepping with a rename of the resource. Since [version 2.9.1](https://github.com/chef-cookbooks/apt/blob/master/CHANGELOG.md#v291) at least which is over a year old so the README can be updated to suggest 2.9.1 or newer for apt.

olivielpeau:
Hi @martinisoft, thanks for the suggestion.

I agree that removing this resource is the better solution, but since it's technically a breaking change and that the current solution works okay on all Chef versions this cookbook supports, I'm going to plan this for the next major bump of the cookbook. Let me know what you think!

martinisoft:
I would agree @olivielpeau so keep me posted about what you'd like to do in that regard. You can combine it with the #450 PR for the major bump.

martinisoft:
I have rebased this against master to get CI builds passing.

martinisoft:
I rebased to sync back up to master so this is ready for review and merge (again)

Please note this will force you to use a slightly newer version of the `apt` cookbook if you're still on an earlier version of chef 12.

martinisoft:
Checking in on this PR again to see if there is any intent to merge. Since I have switched companies and no longer deploy anything Chef I'm not putting a lot of energy into Chef work these days. If there isn't I'll go ahead and close out this PR, if there is I'll make a fresh rebase for review.

olivielpeau:
Hi @martinisoft, thanks for the update!

We do intend to merge this, but for the next major release (now `4.0`). It'll likely take some time for us to get to 4.0 so I don't think you need to rebase this PR, we can take care of that whenever we actually prepare 4.0.

Since it seems we won't see you on this repo much in the future, I'd like to thank you for your significant contributions to this cookbook over the years. I haven't been always very timely in the reviews and releases (to say the least), but have always appreciated your insights and code contributions!

PR #455, opened 2017-07-14
Fix host parameter name in redisdb example

benmanns:


olivielpeau:
The current attribute documentation is actually correct: the attribute key to set is `server`, and the value is then used for the `host` parameter in the yaml template (see https://github.com/DataDog/chef-datadog/blob/v2.11.0/templates/default/redisdb.yaml.erb#L6).

This behavior is quite confusing but changing it would be a breaking change which I don't think is worth making.

I'm going to go ahead and close the PR, but let me know if you have thoughts on this

PR #465, opened 2017-09-12
Add configuration for the process-agent

conorbranagan:
Adds the flag for toggling the process-agent plus some additional optional configuration around the process-agent behavior.


olivielpeau:
@conorbranagan Actually I realize now that the attribute  key `node['datadog']['process']` is already used in this cookbook to configure the process check (see https://github.com/DataDog/chef-datadog/blob/v2.10.1/recipes/process.rb#L28).

What do you think of renaming the `process` attributes to `process_agent` to avoid any confusion?

PR #484, opened 2017-10-15
Fixed #35 - Add recipe for go-metro integration

iancward:
The recipe will:
- handle installing the desired package for libcap
- allow a user to override the package_name and any property of the libcap package resource
- configure and deploy the go-metro yaml config


iancward:
@olivielpeau any updates or feedback on this?

Thanks.

rmoriz:
Using a dash "-" in the attribute name was a very poor decision, because even simple erb/template things explode e.g.:

```ruby
template '/root/.dogrc' do
  source 'dogrc.erb'
  variables(
    node['datadog']
  )
  mode 0o600
end
```

template:

```

[Connection]
apikey: <%= @api_key %>
appkey: <%= @application_key %>
```
==>

```
    
    NameError
    ---------
    `@go-metro' is not allowed as an instance variable name
```

PR #496, opened 2017-12-05
[hybrid] adding hybrid mode for agents - do not merge to master

truthbk:


libraries/recipe_helpers.rb
```diff
@@ -10,6 +10,17 @@ def application_key(node)
         run_state_or_attribute(node, 'application_key')
       end
 
+      def is_agent6(node)
+        if run_state_or_attribute(node, 'agent6') == true
+          if run_state_or_attribute(node, 'hybrid') == true
```
truthbk: I know this isn't very idiomatic, but surprisingly, `run_state_or_attribute` was still truthy despite the attribute being `false`

recipes/dd-agent.rb
```diff
@@ -44,7 +44,10 @@
 # To add integration-specific configurations, add 'datadog::config_name' to
 # the node's run_list and set the relevant attributes
 #
-if node['datadog']['agent6']
+Chef::Log.warn("node[datadog][agent6] set to #{node['datadog']['agent6']}")
+Chef::Log.warn("node[datadog][hybrid] set to #{node['datadog']['hybrid']}")
+Chef::Log.warn("node computed to be agent6: #{Chef::Datadog.agent6?(node)}")
```
olivielpeau: I'm okay with logging this, but could we use `info`-level instead? :)

truthbk:
We will finally not be implementing the hybrid role on `production` so closing this PR. Thanks for the reviews @olivielpeau 🙇

PR #497, opened 2017-12-05
Fix datadog apt key check

bai:
_(Updated PR description)_

`not_if` condition has been failing because of an extra space in key signature, forcing chef to re-fetch key from keyserver on every run.

```
$ apt-key list
/etc/apt/trusted.gpg
--------------------
pub   rsa4096 2016-06-29 [SC] [expires: 2022-06-28]
      A292 3DFF 56ED A6E7 6E55  E492 D3A8 0E30 382E 94DE
uid           [ unknown] Datadog, Inc <package@datadoghq.com>
sub   rsa4096 2016-06-29 [S] [expires: 2022-06-28]
sub   rsa4096 2016-06-29 [S] [expires: 2022-06-28]
sub   rsa4096 2016-07-11 [S] [expires: 2022-07-10]
```

cc: @olivielpeau

### DIFF
--- a/recipes/repository.rb
+++ b/recipes/repository.rb
@@ -29,7 +29,7 @@ when 'debian'
   # Trust new APT key
   execute 'apt-key import key 382E94DE' do
     command 'apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 A2923DFF56EDA6E76E55E492D3A80E30382E94DE'
-    not_if 'apt-key list | grep 382E94DE'
+    not_if 'apt-key list | grep "382E 94DE"'
   end
 
   # Add APT repository
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -27,7 +27,7 @@ RSpec.configure do |config|
     stub_command('apt-cache policy datadog-agent-base | grep "Installed: (none)"').and_return(false)
 
     # recipes/repository.rb
-    stub_command('apt-key list | grep 382E94DE').and_return(false)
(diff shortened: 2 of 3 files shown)


recipes/repository.rb
```diff
@@ -29,7 +29,7 @@
   # Trust new APT key
   execute 'apt-key import key 382E94DE' do
     command 'apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 A2923DFF56EDA6E76E55E492D3A80E30382E94DE'
-    not_if 'apt-key list | grep 382E94DE'
+    not_if 'apt-key list | grep "382E 94DE"'
```
olivielpeau: I think the output of `apt-key list` isn't consistent between versions of `apt`. For instance, on an ubuntu 14.04 VM with `apt` version `1.0.1ubuntu2`, the output is:

```
/etc/apt/trusted.gpg
--------------------

pub   4096R/382E94DE 2016-06-29 [expires: 2022-06-28]
uid                  Datadog, Inc <package@datadoghq.com>
sub   4096R/F432F6E0 2016-06-29 [expires: 2022-06-28]
sub   4096R/8387EEAF 2016-06-29 [expires: 2022-06-28]
sub   4096R/BFF6291E 2016-07-11 [expires: 2022-07-10]
```

So we may want to rely on the output of `apt-key fingerprint` instead since its output might be more stable across `apt` versions.

olivielpeau:
Closing since the key of the APT repo has been rotated, and that additional key import removed (the key is now imported directly by the `apt_repository` resource as the repo's main key).

PR #511, opened 2018-02-13
Support process-agent configuration in datadog.yaml

conorbranagan:
We can do this as of this change being merged: https://github.com/DataDog/datadog-agent/pull/1136

The configuration format is defined in https://github.com/DataDog/datadog-process-agent/blob/master/config/yaml_config.go


attributes/default.rb
```diff
@@ -342,31 +342,24 @@
...
-# Example: 'my-secret-app,dbpass'
-default['datadog']['process_agent']['blacklist'] = nil
-
-# Comma-separated list of regex patterns of containers to include or skip.
-# Each pattern should be in the form of "field:pattern" where 'field' is either
-# 'image' or 'name'.
-# Example: 'image:redis,image:nginx'
-default['datadog']['process_agent']['container_blacklist'] = nil
-# Whitelist is applied after the blacklist.
-default['datadog']['process_agent']['container_whitelist'] = nil
```
conorbranagan: Blacklists for containers are handled in the standard Agent 6 way as we share the tagger code.

attributes/default.rb
```diff
@@ -342,31 +342,24 @@
...
 # overrides the collection intervals for the full and real-time checks in seconds.
-default['datadog']['process_agent']['process_interval'] = nil
-default['datadog']['process_agent']['rtprocess_interval'] = nil
+default['datadog']['process_agent']['process_interval'] = 10
+default['datadog']['process_agent']['rtprocess_interval'] = 2
 
 # If only collecting containers ('enable_process_agent' is false but docker is available)
 # overrides the collection intervals for the full and real-time check.
-default['datadog']['process_agent']['container_interval'] = nil
-default['datadog']['process_agent']['rtcontainer_interval'] = nil
```
conorbranagan: we're setting the defaults here explicitly so we don't have to worry about nil values in the config

attributes/default.rb
sunhay: We should also at some point clean up the config blacklist code in the process-agent since it looks like it's never used: https://github.com/DataDog/datadog-process-agent/blob/master/config/config.go#L218

templates/default/datadog.yaml.erb
```diff
@@ -82,6 +93,20 @@ agent_config = @extra_config.merge({
 
   # log agent options
   log_enabled: node['datadog']['enable_logs_agent']
+
+  # process agent options
+  process_dd_url: node['datadog']['process_agent']['url'],
```
truthbk: Note: raclette has put there endpoint within the `apm_config` section. Could we try to be consistent here? Would `dd_url` under `process_config` be better? I know this would imply changes to the process-agent, hopefully it's not too late.

templates/default/datadog.yaml.erb
```diff
@@ -82,6 +93,20 @@ agent_config = @extra_config.merge({
 
   # log agent options
   log_enabled: node['datadog']['enable_logs_agent']
+
```
truthbk: we're missing a comma in the line above :)

templates/default/datadog.yaml.erb
conorbranagan: This is a good idea. I have a PR on the process-agent side in https://github.com/DataDog/datadog-process-agent/pull/109 that will hopefully make it into the RC1

attributes/default.rb
olivielpeau: alternatively, you could remove the keys that have `nil` values in the template logic, with something like `agent_config[:process_config].reject!{ |k,v| v.nil? }`.

Having the defaults defined in the attributes instead of the agent has been a bit of a pain to work with in the past, hence the suggestion

recipes/_agent6_config.rb
```diff
@@ -53,34 +52,6 @@ def conf_template_vars
   notifies :restart, 'service[datadog-agent]', :delayed unless node['datadog']['agent_start'] == false
 end
 
-template process_agent_config_file do
```
olivielpeau: If this code worked for the beta A6 builds, could you keep it (along with the resource `datadog_monitor 'process_agent'`) for now? I can take care of removing it when the Agent 6 is actually GA.

If we remove it now there's no "safe" path to upgrade from A6 `beta.9` to `rc.1` without breaking the process agent, as no version of this cookbook would fully support both.

I'd like the next version of the cookbook (`2.14.0`) to support both `beta.9` and the RCs, and when we make A6 GA I'll release v`3.0.0` of the cookbook with A6 installed by default, and all this legacy code removed.

conorbranagan:
@olivielpeau Should be good now, thanks

PR #525, opened 2018-03-05
Set windows values for agent 6

### LINKED ISSUE #524: Agent 6 doesn't install on Windows
There are several issues with installing Agent 6 on Windows
1. Dependent service restart (fixed by #520 )
1. Config file security not setting Windows values
1. No default for Windows config directory

The config directory for integrations has also changed for Agent 6 to be a sub directory of conf.d (see https://github.com/DataDog/datadog-agent/blob/master/docs/agent/changes.md#checks-configuration-files)

rlaveycal:
- permission config file on windows
- install checks yaml in sub dir of conf.d on agent 6

Closes #524

### DIFF
--- a/providers/monitor.rb
+++ b/providers/monitor.rb
@@ -75,7 +75,7 @@ private
 
 def yaml_dir
   if node['datadog']['agent6']
-    ::File.join(node['datadog']['agent6_config_dir'], 'conf.d', "#{new_resource.name}.d")
+    ::File.join(node['datadog']['agent6_config_dir'], 'conf.d')
   else
     ::File.join(node['datadog']['config_dir'], 'conf.d')
   end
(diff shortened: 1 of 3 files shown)


providers/monitor.rb
```diff
@@ -72,3 +70,13 @@ def whyrun_supported?
     provider service_provider unless service_provider.nil?
   end
 end
+
+private
+
+def yaml_dir
+  if node['datadog']['agent6']
+    ::File.join(node['datadog']['agent6_config_dir'], 'conf.d', "#{new_resource.name}.d")
```
olivielpeau: could you remove the `"#{new_resource.name}.d"` part to keep only `::File.join(node['datadog']['agent6_config_dir'], 'conf.d')`? Agent 6 still supports that configuration scheme (and that's the scheme the cookbook currently uses for Linux Agent 6). Changing the scheme now would slightly break the cookbook for Linux Agent 6 users upgrading from previous versions of the cookbook.

I'm planning to switch to the new scheme in the next major version of the cookbook.

PR #533, opened 2018-03-23
[docs] Document easy_install_package removal from Chef 13

olivielpeau:
Affects the dogstatsd-python and ddtrace-python recipes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,9 +95,23 @@ dogstatsd-(python|ruby)
 -----------------------
 Installs the language-specific libraries to interact with `dogstatsd`.
 
+* Note for Chef >= 13 users: the `datadog::dogstatsd-python` recipe is not compatible with Chef >= 13, as it relies on a resource that was removed in Chef 13.0.
+  To install the `dogstatsd-python` library with Chef, please add a dependency on the `poise-python` cookbook to your custom/wrapper cookbook, and use the following resource:
+  ```ruby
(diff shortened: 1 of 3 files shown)


haidangwa:
Can we also get active code in place with this PR to detect chef versions >= 13 and throw a warning into the client.log? I think the documentation effort is a good start, but in-code, this needs to be a bit more active.

Do you see any functional downsides to using ddtrace-ruby and dogstatsd-ruby.rb in place of their python counterparts?

haidangwa:
Maybe something like this:

olivielpeau:
Thanks for the feedback @haidangwa! I prefer failing early than ignoring the error and printing a warning, let me explain why:

Chef 12 itself deprecates `easy_install_package` by printing this deprecation warning when the resource is used: `The easy_install package provider is deprecated and will be removed in Chef 13. at 1 location: [...]`. To me this should cover users currently using Chef 12 and this recipe, and who're planning an upgrade to 13.

For Chef 13 users that would attempt using the recipe, the chef run would fail at compile time with this error:
```
NoMethodError
       -------------
       undefined method `easy_install_package' for cookbook: datadog, recipe: dogstatsd-python :Chef::Recipe
```

To me failing with such a clear error is better than the recipe not doing what it's supposed to do (and printing a warning). This is also the way Chef itself deprecates its own features.

PR #538, opened 2018-04-09
Removed default values for resource properties

stefanwb:
This eliminates the warnings when no value was provided that concerns
different behaviour depending in Chef 12 than in Chef 13

Signed-off-by: Stefan Wessels Beljaars <swesselsbeljaars@schubergphilis.com>


AliDatadog:
Closed for inactivity.

PR #540, opened 2018-04-16
Moises.botarro/adding data scrubber fields process agent

moisesbotarro:
Adding new fields for process-agent to enable/disable the cmdline data scrubber and to expand list of sensitive words to match.
@DataDog/burrito


attributes/default.rb
```diff
@@ -367,6 +367,11 @@
 # Example: ['my-secret-app', 'dbpass']
 default['datadog']['process_agent']['blacklist'] = []
 
+# Controls the behavior of the cmdline dataScrubber
+# Example for custom sensitive words: ['consul_token', 'token', 'dd_api_key']
```
conorbranagan: maybe put this comment above the `custom_sensitive_words` and instead add a comment here on what `scrub_args` enables?

PR #544, opened 2018-05-10
Update default.rb - typo fix

rsheyd:
type for enabling logs agent in the comment
where the line is checked: https://github.com/DataDog/chef-datadog/blob/master/templates/default/datadog.yaml.erb#L90


gwaldo:
Thank you, @rsheyd

PR #546, opened 2018-06-22
add support for dogstatsd_socket tunable in datadog.yaml

Klintrup:
add support for dogstatsd_socket setting in datadog.yaml https://docs.datadoghq.com/developers/dogstatsd/unix_socket/

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -70,6 +70,8 @@ default['datadog']['agent6_config_dir'] =
 # Leave key value to nil to use agent6 default for that endpoint.
 # Supported keys: "series", "events", "service checks"
 default['datadog']['use_v2_api'] = {}
+# set path to dogstatsd unix socket - https://docs.datadoghq.com/developers/dogstatsd/unix_socket/
+default['datadog']['dogstatsd_socket'] = false
 
 ###                 End of Agent6-only attributes                    ###
 ########################################################################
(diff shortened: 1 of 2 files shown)


templates/default/datadog.yaml.erb
```diff
@@ -112,6 +112,10 @@ if node['datadog']['statsd_forward_host']
   agent_config['statsd_forward_port'] = node['datadog']['statsd_forward_port']
 end
 
+if node['datadog']['agent6'] && node['datadog']['dogstatsd_socket']
+  agent_config['dogstatsd_socket'] = node['datadog']['dogstatsd_socket']
```
olivielpeau: if the default is left to `nil`, you can move set `dogstatsd_socket` directly in the block line 72 like the other regular config options. `nil` values are then automatically removed from the hash.

attributes/default.rb
```diff
@@ -70,6 +70,8 @@
 # Leave key value to nil to use agent6 default for that endpoint.
 # Supported keys: "series", "events", "service checks"
 default['datadog']['use_v2_api'] = {}
+# set path to dogstatsd unix socket - https://docs.datadoghq.com/developers/dogstatsd/unix_socket/
+default['datadog']['dogstatsd_socket'] = false
```
olivielpeau: Can you leave the default value to `nil` here? In general for agent options we prefer letting the Agent determine what the correct default instead of the cookbook.

PR #547, opened 2018-07-05
[circleci] migrating to v2

truthbk:


circle.yml
```diff
@@ -1,26 +0,0 @@
...
-    - "~/.chefdk"
-
-test:
-  pre:
-    - eval "$(/opt/chefdk/bin/chef shell-init bash)"
-  override:
-    - chef --version
-    - chef gem query -d chefspec rubocop foodcritic rake kitchen-docker
-    - chef exec rake circle:
-        parallel: true
```
hush-hush: Do we need to add `parallel` and `timeout` to the new circleci file ?

truthbk: I couldn't find the way to add the directives to the job in `v2`. I'll take another look.

hush-hush: OK, we should just double check what they did in V1 and that we no longer need them with V2.

truthbk: setting parallelism would not help in circleci v2. It didn't really accomplish much in v1 anyways.

PR #548, opened 2018-07-11
instance level min_collection_interval

mhebbar1:
Instance level min_collection_interval is supported, but was not used in the template file. Hence adding it.


remeh:
Thanks for your contribution @mhebbar1 

I'm merging it, it'll be part of `2.17.0`

PR #549, opened 2018-07-13
add no_proxy option

stonith:
TL:DR Adds no_proxy option

The behaviour in agent v6 seems to behave differently than v5 when a web proxy is configured. With v5, the agent would connect directly to 127.0.0.1 whereas with v6, the agent will try to connect to 127.0.0.1 via the proxy. In v6, there is a no_proxy option which is not supported the chef datadog cookbook.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -296,7 +296,7 @@ default['datadog']['web_proxy']['port'] = nil
 default['datadog']['web_proxy']['user'] = nil
 default['datadog']['web_proxy']['password'] = nil
 default['datadog']['web_proxy']['skip_ssl_validation'] = nil # accepted values 'yes' or 'no'
-default['datadog']['web_proxy']['no_proxy'] = nil
+default['datadog']['web_proxy']['no_proxy'] = nil # only used for agent v6.0+
 
 # dogstatsd
 default['datadog']['dogstatsd'] = true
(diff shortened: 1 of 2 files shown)


attributes/default.rb
```diff
@@ -296,6 +296,7 @@
 default['datadog']['web_proxy']['user'] = nil
 default['datadog']['web_proxy']['password'] = nil
 default['datadog']['web_proxy']['skip_ssl_validation'] = nil # accepted values 'yes' or 'no'
+default['datadog']['web_proxy']['no_proxy'] = nil
```
olivielpeau: Could you add a small comment here to mention it's used for Agent v6+ only?

templates/default/datadog.yaml.erb
```diff
@@ -37,6 +37,7 @@ if node['datadog']['web_proxy']['host']
   user = node['datadog']['web_proxy']['user']
   password = node['datadog']['web_proxy']['password']
   scheme = ""
+  no_proxy = node['datadog']['web_proxy']['no_proxy']
```
olivielpeau: can you initialize `no_proxy` to nil outside of this `if` block

templates/default/datadog.yaml.erb
```diff
@@ -136,6 +137,10 @@ if !http_proxy.nil?
   }
 end
 
+if !no_proxy.nil?
+  agent_config['proxy']['no_proxy'] = no_proxy
```
olivielpeau: `agent_config['proxy']` should be initialized to a hash here, but the logic is a bit brittle. Could you either check that the key is indeed present in `agent_config` or move this block inside the `!http_proxy.nil?` conditional above?

pallan:
@olivielpeau comments addressed. Thanks!

Issue #542, opened 2018-04-23
Problem installing agent6 on ubuntu 14.04 due to upstart


sylvaintremblay666:
Installing agent6 on ubuntu 14.04 doesn't work due to upstart.

A bug have already been fixed regarding this for RHEL / Amazon Linux on version 2.14.0 / 2018-02-27, but it doesn't include Ubuntu.

In _recipies/dd-agent.rb_ around line 106, a case for ubuntu 14.04 should be added. To test locally, I simply forced "service_provider = Chef::Provider::Service::Upstart" which made the installation succeed.

olivielpeau:
Hi @sylvaintremblay666, thanks for opening this issue!

Chef should detect `upstart` as the default service provider on Ubuntu 14.04 (as upstart _is_ the default init system on Ubuntu 14.04). What's your version of Chef? Is there anything specific about your init system setup on your Ubuntu 14.04 environment?

sylvaintremblay666:
Hi @olivielpeau ,

I'm running on AWS with OpsWorks. Chef is a little old (11.10.4).
Running on this AMI : ubuntu/images/hvm/ubuntu-trusty-14.04-amd64-server-20170811 (ami-845367ff)
Nothing special about the machine / its init system.

Thanks a lot,

Sylvain

mshepet:
+1
I ran into the exact same issue with agent6 on OpsWorks w/ Chef 11.10 and Ubuntu 14.04. Used sylvaintremblay666's solution to force service_provider = Chef::Provider::Service::Upstart in dd-agent.rb and that solved the installation error.

remeh:
Hello! Indeed, Chef < 11.14 doesn't use the correct service provider for Ubuntu. This PR has fixed it. https://github.com/DataDog/chef-datadog/pull/551 and it'll be available in release `2.18.0`.

PR #551, opened 2018-07-21
Forces ubuntu 14.04 on Chef 11 to use upstart

rposborne:
This addresses https://github.com/DataDog/chef-datadog/issues/542

### DIFF
--- a/recipes/dd-agent.rb
+++ b/recipes/dd-agent.rb
@@ -105,7 +105,7 @@ end
 service_provider = nil
 if node['datadog']['agent6'] &&
    (((node['platform'] == 'amazon' || node['platform_family'] == 'amazon') && node['platform_version'].to_i != 2) ||
-   (node['platform'] == 'ubuntu' && node['platform_version'].to_f == 14.04) ||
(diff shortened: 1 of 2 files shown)


recipes/dd-agent.rb
```diff
@@ -105,6 +105,7 @@ def template_vars # rubocop:disable Metrics/AbcSize
 service_provider = nil
 if node['datadog']['agent6'] &&
    (((node['platform'] == 'amazon' || node['platform_family'] == 'amazon') && node['platform_version'].to_i != 2) ||
+   (node['platform'] == 'ubuntu' && node['platform_version'].to_f == 14.04) ||
```
olivielpeau: * Unfortunately this logic is currently duplicated in `https://github.com/DataDog/chef-datadog/blob/master/providers/monitor.rb` (twice), so you also need to update it there.
* Ubuntu switched to systemd in `15.04`, what do you think of comparing with `< 15.04` instead? (should work since we're comparing floats)

remeh:
Thank you for your contribution @rposborne, I'm merging it and it will be part of release `2.18.0`.

rposborne:
@remeh Thanks for cleaning up after me!!

PR #558, opened 2018-09-10
Fixed #550: Handle agent v6 integration conf files

### LINKED ISSUE #550: configuration file locations is not up-to-date with default installation in agent 6
When configuring a redisdb monitor in the datadog cookbook, the cookbook creates file 
`/etc/datadog-agent/conf.d/redisdb.yaml`.  In agent version 6, the path is supposed to be 
`/etc/datadog-agent/conf.d/redisdb.d/conf.yaml`.  

It does not appear that there is code to differentiate this location change (I'm assuming such code would live approximately [here](https://github.com/DataDog/chef-datadog/blob/b3cf08abcb09f3656dca71524e1bc0ee7c625eef/providers/monitor.rb#L12)?)

Are there plans to fix where the yaml files are populated for the new layout?

iancward:


providers/monitor.rb
```diff
@@ -77,10 +76,12 @@ def whyrun_supported?
...
-def yaml_dir
+def conf_yaml_path(resource_name)
   if node['datadog']['agent6']
-    ::File.join(node['datadog']['agent6_config_dir'], 'conf.d')
+    ::File.join(node['datadog']['config_dir'], 'conf.d',
+                "#{resource_name}.d", 'conf.yaml')
   else
-    ::File.join(node['datadog']['config_dir'], 'conf.d')
+    ::File.join(node['datadog']['agent6_config_dir'], 'conf.d',
+                "#{resource_name}.yaml")
```
jeffbyrnes: I think the `config_dir` attributes were reverse accidentally here. I think this should be:

```ruby
if node['datadog']['agent6']
  ::File.join(
    node['datadog']['agent6_config_dir'],
    'conf.d',
    "#{resource_name}.d",
    'conf.yaml'
  )
else
  ::File.join(
    node['datadog']['config_dir'],
    'conf.d',
    "#{resource_name}.yaml"
  )
end
```

jeffbyrnes: I.e., `git revert 68d00c8463f0144cba6a1186e31ab7c652888362`

iancward: Ah, it looks like in 68d00c8463f0144cba6a1186e31ab7c652888362 I switched both the base agent config path as well as the resource path; I should have only switched the resource path. So we don't want a full revert. Fixed in 167a877.

olivielpeau:
Thanks @iancward for the PR, and @jeffbyrnes for the review.

I feel this change could be more graceful for existing users, especially since this path change isn't required by Agent 6: with this change, existing Agent 6 users who upgrade their cookbook will end up with both `/etc/datadog-agent/conf.d/<integration_name>.yaml` and `/etc/datadog-agent/conf.d/<integration_name>.d/conf.yaml`, and twice the number of configured check instances they configured, on the nodes that were previously provisioned with the previous version of the cookbook.

What do you think of (when Agent 6 is installed) making the provider remove the file at `/etc/datadog-agent/conf.d/<integration_name>.yaml`, unless an attribute is set on the resource to disable this?

iancward:
@olivielpeau agent v6 changed the config file location from `/etc/dd-agent/conf.d/<integration_name.yaml>` to `/etc/datadog-agent/conf.d/<integration_name>.d/conf.yaml`. The current (released) version of the cookbook still writes files to the old dd-agent directory, which the v6 agent doesn't even look at. So as it stands, this cookbook can't be used to configure usable integrations for agent v6 at all (and the reason why I'm currently unable to move to agent v6).
That is to say, if someone is using this cookbook to deploy agent v6, and has attempted to use it to also manage configurations, then chef is writing files that will never be looked at. I would assume that end-users would have already noticed this and either reverted back to agent v5 or found other means to configure their integrations. 

If someone is moving from agent v5 to agent v6, I believe the migration install process should automatically handle migrating the integration files to the right place, and on subsequent runs, with this proposed change, chef would go from managing the old file(s) to managing the new ones. I'm not certain if the migration process deletes the old configuration files.

All this being said, since the current state is "broken", I'm not sure if there's value in trying to clean up the old files. Going the proposed route would also only delete the agent integrations that chef has been told to manage.

jeffbyrnes:
@iancward so the Agent v6 _does_ look at the old location; if it did not, a bunch of my configs would have stopped working some time ago.

So I think what @olivielpeau recommends has merit!

olivielpeau:
@iancward To clarify, there are 2 distinct changes that Agent v6 introduced on the integration config files:
1. Breaking change: on Linux, the root of the config folder in v6 is `/etc/datadog-agent/` instead of `/etc/dd-agent`. This breaking change is currently handled transparently by the cookbook, see for example https://github.com/DataDog/chef-datadog/blob/v2.16.1/providers/monitor.rb#L80-L86
2. Backward-compatible change (new feature): for a given integration `foo`, the Agent v6 looks for configuration files in  `<root_conf_dir>/conf.d/foo.d/*.yaml` _in addition to_ `<root_conf_dir>/conf.d/foo.yaml`. At the moment the cookbook still writes to `<root_conf_dir>/conf.d/foo.yaml`, but that's fully compatible with Agent v6.

See also our docs on this: https://github.com/DataDog/datadog-agent/blob/master/docs/agent/changes.md#checks-configuration-files.

Let me know if anything is unclear, or if you encounter other issues upgrading to Agent 6 with this cookbook.

So the present PR changes are not _required_ for Agent 6 to work properly.

There's a feature we could definitely add to the cookbook though: the ability for the cookbook to write to a `<root_conf_dir>/conf.d/foo.d/<arbitrary_name>.yaml` file, which would bring more flexibility in the usage of the `datadog_monitor` resource. For example, `datadog_monitor` would be able to be called multiple times for the same integration, and would create different conf files for this integration, which is not possible with the resource at the moment.

dfernandoDev:
Any update on this pull request?

joshuawscott:
Any idea on when this might be merged? This would be super helpful for our use cases, especially in monitoring processes and redis instances.

Thanks!

albertvaka:
We finally implemented this feature in https://github.com/DataDog/chef-datadog/pull/666 which will be released as version 3.5.0. Can you guys give it a try?

Thanks a lot for your patience and special thanks to @iancward for working on this PR :)

PR #561, opened 2018-09-17
Retry on failure when pulling the gpg key from keyserver.ubuntu.com

remicalixte:
# Motivation

the keyservers are a pool of sks servers with squid in front of them, we should probably retry on failures.

### DIFF
--- a/recipes/repository.rb
+++ b/recipes/repository.rb
@@ -37,6 +37,7 @@ when 'debian'
     distribution distribution
     components components
     action :add
+    retries 4
   end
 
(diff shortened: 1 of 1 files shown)


recipes/repository.rb
```diff
@@ -37,6 +37,7 @@
     distribution distribution
     components components
     action :add
+    retries 4
```
olivielpeau: Could you make the number of retries configurable? (a default of `4` sounds good)

remicalixte: Yes I'll add a parameter

PR #562, opened 2018-09-19
add support for setting max_detailed_exchanges for RabbitMQ

asherf:
RabbitMQ integration comes with a limit of 50 exchanges. 
This is to support more than 50 exchanges per RMQ node.
it will impact agent's perf.


remeh:
Thanks for your contribution @asherf 
It'll be part of the next release!

PR #565, opened 2018-10-10
Switch to using cookstyle

jeffbyrnes:
Maintain the current style choices, but does away with a lot of unnecessary rule bending/exclusion.


olivielpeau:
Awesome, thanks for the quick fix! Merging this 👍

Issue #541, opened 2018-04-16
Foodcritic errors


jjasghar:
```
FC079: Deprecated easy_install_package resource usage: ./recipes/ddtrace-python.rb:9
FC113: Resource declares deprecated use_inline_resources: ./providers/monitor.rb:4
```
https://github.com/DataDog/chef-datadog/issues/539

Can yall either accept these style differences and ignore them or resolve it?

/cc @russellseymour

jjasghar:
```
[datadog-cookbook] $ /bin/bash /tmp/jenkins7673545352440283996.sh
Checking 144 files
........x....................................x..................................................................................................
FC079: Deprecated easy_install_package resource usage: ./recipes/ddtrace-python.rb:9
FC113: Resource declares deprecated use_inline_resources: ./providers/monitor.rb:4
Build step 'Execute shell' marked build as failure
Finished: FAILURE
```

PR #567, opened 2018-10-10
Replace easy_install resource with python_package

### LINKED ISSUE #539: Replace easy_install resource with something Chef 13 proof
Deprecation: Easy Install Resource (CHEF-6)

The Python community recommends that users prefer pip rather than easy_install to install python packages.

The easy_install resource was deprecated in 12.10, and will be removed in Chef 13.

Remediation
There is no built in replacement for easy_install in Chef. The poise-python cookbook provides a set of resources for managing Python installations.

https://docs.chef.io/deprecations_easy_install.html

jeffbyrnes:
Adds a dependency on poise-python.

Fixes #539

Relates to #541

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,12 +106,7 @@ ddtrace-(python|ruby)
 ---------------------
 Installs the language-specific libraries for application Traces (APM).
 
-* Note for Chef >= 13 users: the `datadog::ddtrace-python` recipe is not compatible with Chef >= 13, as it relies on a resource that was removed in Chef 13.0.
-  To install the `ddtrace-python` library with Chef, please add a dependency on the `poise-python` cookbook to your custom/wrapper cookbook, and use the following resource:
-  ```ruby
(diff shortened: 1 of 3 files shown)


olivielpeau:
I'd like to avoid adding a dependency on a new cookbook in order to only provide this recipe: 

* maintaining cookbook dependencies/version constraints over time is always a bit tricky (for example see all the "recommendation" notes in `metadata.rb` and in the README), and
* I feel the use case of installing the `ddtrace` pip package could be more easily addressed by documenting it in the README

Instead, I'd prefer deprecating the `ddtrace-python` recipe in `datadog` `2.17` , and remove it in `3.0` (while documenting how to use `poise-python` to achieve the same purpose). What do you think?

jeffbyrnes:
@olivielpeau I think that’s reasonable! I think it’s probably out-of-scope for me to tackle a deprecation in a cookbook I don’t maintain directly, so I’ll close this & leave that to you.

PR #569, opened 2018-10-22
Fixes an issue with RHEL 7+

Nibons:
Issue happened when trying to use a rhel7+ node, or centos7+

rhel7+ uses a different mechanism for controlling services (Chef::Provider::Service::Systemd)

### DIFF
--- a/recipes/dd-agent.rb
+++ b/recipes/dd-agent.rb
@@ -102,12 +102,13 @@ else
 end
 
 # Common configuration
-service_provider = nil
 if node['datadog']['agent6'] &&
-   (((node['platform'] == 'amazon' || node['platform_family'] == 'amazon') && node['platform_version'].to_i != 2) ||
-   (node['platform'] != 'amazon' && node['platform_family'] == 'rhel' && node['platform_version'].to_i < 7))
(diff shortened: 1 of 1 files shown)


recipes/dd-agent.rb
```diff
@@ -102,12 +102,13 @@ def template_vars # rubocop:disable Metrics/AbcSize
...
 # Common configuration
-service_provider = nil
 if node['datadog']['agent6'] &&
-   (((node['platform'] == 'amazon' || node['platform_family'] == 'amazon') && node['platform_version'].to_i != 2) ||
-   (node['platform'] != 'amazon' && node['platform_family'] == 'rhel' && node['platform_version'].to_i < 7))
+  (((node['platform'] == 'amazon' || node['platform_family'] == 'amazon') && node['platform_version'].to_i != 2) ||
+  (node['platform'] != 'amazon' && node['platform_family'] == 'rhel' && node['platform_version'].to_i < 7))
   # use Upstart provider explicitly for Agent 6 on Amazon Linux < 2.0 and RHEL < 7
   service_provider = Chef::Provider::Service::Upstart
+else
```
olivielpeau: this `else` block will also include Agent 5, but with Agent 5 we should never hardcode the `service_provider` (Agent 5 generally only provides sysv-init scripts, and relies on the init system's compatibility layer with sysv-init scripts)

PR #575, opened 2018-12-03
Add domain socket support to postgres monitor

emillbrandt:
Add an example.
Do not set 'port' in the config when using a domain socket.


remeh:
Hello @emillbrandt !
Thank you for your contribution, sorry I took this long to look at it.
LGTM, it could change the generated configuration for some users but as the port is most likely not needed to use the domain socket, would not have any impact 👌 
Merging!

PR #577, opened 2018-12-10
Optional cmd_port & GUI_port on agentv6 due to port conflicts

MCKrab:
https://github.com/DataDog/datadog-agent/blob/master/pkg/config/config_template.yaml


remeh:
Thank you for your contribution @MCKrab , it has been merged as part of #632 !

PR #582, opened 2019-02-18
Add support for 'site' option.

remeh:
Agent 6.9.0 has a new option `site` to configure on which site (US/EU) the Agent data should be sent, simplifying the configuration process by avoiding to type complete URLs in the `url` configuration field.

This pull-requests adds the attribute `site` in the Chef cookbook in order to use this option by default instead of the `url` one.

Note that it doesn't remove anything about the `url` configuration field has it's still supported and has priority on the `site` value.

The use of `DATADOG_HOST` has been removed: it was used by the chef-handler-datadog recipe to set the datadog host to use, however, the ruby lib dogapi-rb used already has a correct default value if `DATADOG_HOST` was not set and on top of that: https://github.com/DataDog/chef-handler-datadog/pull/103 makes sure that a default value is set directly inside the recipe instead of in the ruby lib dogapi-rb.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -96,13 +96,10 @@ default['datadog']['tag_prefix'] = 'tag:'
 # Don't change these.
 # The host of the Datadog intake server to send Agent data to, only set this option
 # if you need the Agent to send data to a custom URL.
-# For Agent 6, defaults to nil because "url" overrides the site setting defined in "site".
-default['datadog']['url'] =
-  if node['datadog']['agent6'] == true
-    nil
-  else
-    'https://app.datadoghq.com'
-  end
+# The nil value will let the Agent 6 select the URL to send the data.
+# For Agent 5, the Agent 5 recipe will fallback on https://app.datadoghq.com
+# (see recipes/dd-agent.rb).
+default['datadog']['url'] = nil
 
 # Add tags as override attributes in your role
 # This can be a string of comma separated tags or a hash in this format:
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -66,6 +66,10 @@ default['datadog']['agent6_config_dir'] =
     '/etc/datadog-agent'
   end
 
+# The site of the Datadog intake to send Agent data to.
+# Defaults to 'datadoghq.com', set to 'datadoghq.eu' to send data to the EU site.
+default['datadog']['site'] = 'datadoghq.com'
+
 # Set a key to true to make the agent6 use the v2 api on that endpoint, false otherwise.
 # Leave key value to nil to use agent6 default for that endpoint.
 # Supported keys: "series", "events", "service checks"
@@ -88,9 +92,16 @@ default['datadog']['extra_endpoints']['prod']['url'] = nil # optional
 # Set prefix to '' if you want Chef tags to be sent without prefix.
 default['datadog']['tag_prefix'] = 'tag:'
 
-# Don't change these
-# The host of the Datadog intake server to send agent data to
-default['datadog']['url'] = 'https://app.datadoghq.com'
+# Don't change these.
+# The host of the Datadog intake server to send Agent data to, only set this option
+# if you need the Agent to send data to a custom URL.
+# For Agent 6, defaults to nil because "url" overrides the site setting defined in "site".
+default['datadog']['url'] =
+  if node['datadog']['agent6'] == true
+    nil
+  else
+    'https://app.datadoghq.com'
+  end
 
 # Add tags as override attributes in your role
 # This can be a string of comma separated tags or a hash in this format:
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -93,10 +93,11 @@ default['datadog']['extra_endpoints']['prod']['url'] = nil # optional
 # Set prefix to '' if you want Chef tags to be sent without prefix.
 default['datadog']['tag_prefix'] = 'tag:'
 
-# Don't change these.
 # The host of the Datadog intake server to send Agent data to, only set this option
 # if you need the Agent to send data to a custom URL.
 # The nil value will let the Agent 6 select the URL to send the data.
+# Any non-nil value overrides the 'site' value, prefer using 'site' unless your
+# use case isn't covered by 'site'.
 # For Agent 5, the Agent 5 recipe will fallback on https://app.datadoghq.com
 # (see recipes/dd-agent.rb).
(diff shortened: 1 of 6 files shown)


attributes/default.rb
```diff
@@ -88,9 +92,16 @@
...
-# Don't change these
-# The host of the Datadog intake server to send agent data to
-default['datadog']['url'] = 'https://app.datadoghq.com'
+# Don't change these.
+# The host of the Datadog intake server to send Agent data to, only set this option
+# if you need the Agent to send data to a custom URL.
+# For Agent 6, defaults to nil because "url" overrides the site setting defined in "site".
+default['datadog']['url'] =
+  if node['datadog']['agent6'] == true
+    nil
```
olivielpeau: I think `url` is also used to configure the endpoint of the handler (see https://github.com/DataDog/chef-datadog/blob/v2.16.1/recipes/dd-handler.rb#L71 and https://github.com/DataDog/chef-datadog/blob/v2.16.1/recipes/dd-handler.rb#L30), will this still work fine?

attributes/default.rb
```diff
@@ -66,6 +66,10 @@
     '/etc/datadog-agent'
   end
 
+# The site of the Datadog intake to send Agent data to.
+# Defaults to 'datadoghq.com', set to 'datadoghq.eu' to send data to the EU site.
+default['datadog']['site'] = 'datadoghq.com'
```
olivielpeau: Ideally, we should leave the default to `nil` and only write `site` in the config if it's not `nil`.

This allows the Agent to define what the default should be, instead of the cookbook. This isn't the way most attributes were added to the cookbook (and changing that is a breaking change), but it allows greater flexibility and makes the cookbook easier to maintain.

attributes/default.rb
```diff
@@ -66,6 +66,10 @@
     '/etc/datadog-agent'
   end
 
+# The site of the Datadog intake to send Agent data to.
```
olivielpeau: let's mention here which versions of the Agent actually support this config option

recipes/dd-agent.rb
```diff
@@ -39,6 +39,11 @@
 # Set the correct Agent startup action
 agent_start = node['datadog']['agent_start'] ? :start : :stop
 
+# use either the site option or the url one, the latter having priority.
+node.run_state['dd_url'] = 'https://app.datadoghq.com'
+node.run_state['dd_url'] = 'https://app.' + node['datadog']['site'] unless node['datadog']['site'].nil?
+node.run_state['dd_url'] = node['datadog']['url'] unless node['datadog']['url'].nil?
```
olivielpeau: I'd avoid using the run state (to me it's a code smell, even though I don't have chef code style guidelines to justify this). This logic could live in the function `template_vars` below instead

recipes/dd-handler.rb
```diff
@@ -26,7 +26,13 @@
...
 
 include_recipe 'chef_handler'
-ENV['DATADOG_HOST'] = node['datadog']['url']
+
+# use either the site option or the url one, the latter having priority.
+node.run_state['dd_url'] = 'https://app.datadoghq.com'
+node.run_state['dd_url'] = 'https://app.' + node['datadog']['site'] unless node['datadog']['site'].nil?
+node.run_state['dd_url'] = node['datadog']['url'] unless node['datadog']['url'].nil?
+
+ENV['DATADOG_HOST'] = node.run_state['dd_url']
```
olivielpeau: I think setting `DATADOG_HOST` is needed only until `v0.9.0` of the handler ([`v0.10.0`](https://github.com/DataDog/chef-handler-datadog/releases/tag/v0.10.0)) added support for multiple endpoints, which required supporting the `url` config option directly on the handler)

It could be worth removing support for <v0.10.0 in the next major version of the cookbook. Until then, I'd prefer duplicated logic (and a `TODO` or `FIXME` comment) than storing stuff in the `run_state`

attributes/default.rb
```diff
@@ -88,9 +93,16 @@
...
+# Don't change these.
+# The host of the Datadog intake server to send Agent data to, only set this option
+# if you need the Agent to send data to a custom URL.
+# For Agent 6, defaults to nil because "url" overrides the site setting defined in "site".
+default['datadog']['url'] =
+  if node['datadog']['agent6'] == true
+    nil
+  else
+    'https://app.datadoghq.com'
+  end
```
olivielpeau: I don't think this conditional logic is needed: I don't think you need to set `default['datadog']['url']` to a non-nil value in any case given the logic you've added in the recipes.

Also, this conditional logic will be resolved early at compile time, before all the node attributes are resolved by Chef, so it will very likely give unexpected results in practice if the attribute `['datadog']['agent6']` is set in other places by users.
(in case you're wondering: the other node attributes that we do use in the present file are set by Chef before the compile phase, so they're safe to use here)

remeh: While adding this condition I was indeed wondering if it would be safe to use `node` values here, thanks for the infos. You're right, given the others conditions added in this PR, this one is not useful.

recipes/dd-handler.rb
```diff
@@ -26,7 +26,12 @@
 end
 
 include_recipe 'chef_handler'
-ENV['DATADOG_HOST'] = node['datadog']['url']
+
+if node['datadog']['chef_handler_version'] &&
+   Gem::Version.new(node['datadog']['chef_handler_version']) < Gem::Version.new('0.10.0')
```
olivielpeau: inconsistent indentation

attributes/default.rb
```diff
@@ -88,9 +93,13 @@
 # Set prefix to '' if you want Chef tags to be sent without prefix.
 default['datadog']['tag_prefix'] = 'tag:'
 
-# Don't change these
-# The host of the Datadog intake server to send agent data to
-default['datadog']['url'] = 'https://app.datadoghq.com'
+# Don't change these.
+# The host of the Datadog intake server to send Agent data to, only set this option
+# if you need the Agent to send data to a custom URL.
+# The nil value will let the Agent 6 select the URL to send the data.
```
olivielpeau: Let's add that `Any non-nil value overrides the 'site' value, prefer using 'site' unless your use case isn't covered by 'site'.` (might be a bit long, but I want to encourage users to use `site` instead of `url`, unless they have good reasons to use `url` )

olivielpeau:
also, can you add a short explanation in the PR description about why you can remove `DATADOG_HOST`?

PR #585, opened 2019-03-05
Log deprecation message for recipes using easy_install.

remeh:


recipes/ddtrace-python.rb
```diff
@@ -6,6 +6,11 @@
 # NB: This recipe is compatible with Chef <= 12 only. If you're using Chef 13 or higher
 # please refer to the README of the cookbook.
 
+log 'message' do
+  message 'This recipe will be deprecated in version 3.0.0 of the cookbook. See README.md chapter ddtrace-python for more infos.'
```
olivielpeau: `chapter` -> `section`

`infos` -> `info`

Also, can you remind the name of the cookbook in the message? (`datadog cookbook`). Generally useful.

recipes/ddtrace-python.rb
```diff
@@ -6,6 +6,11 @@
 # NB: This recipe is compatible with Chef <= 12 only. If you're using Chef 13 or higher
 # please refer to the README of the cookbook.
 
+log 'message' do
```
olivielpeau: any reason to use the resource instead of `Chef::Log`? (genuine question, didn't know about the resource)

remeh: Not any, I've changed to use Chef::Log as I see there is already many usage of it in the repo, I prefer to stick with known things.

recipes/ddtrace-python.rb
remeh: thanks 👍 
I've also repeated the recipe name.

PR #588, opened 2019-03-07
Let the Windows installer set the correct rights on config files and directories.

remeh:
The .MSI to setup the Agent on Windows will soon run the Agent in a custom user with less privileges (than Administrator).

This commit removes the permissions change on `datadog.yaml` on Windows to let the installer set it itself.

I've tested that the Agent still runs while installing `6.8.0` and `6.10.0`.

Note that this PR doesn't change the behavior for the Agent 5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,19 @@ Should you wish to add additional elements to the agent6 configuration file (typ
 
 For general information on the Datadog Agent 6, please refer to the [datadog-agent](https://github.com/DataDog/datadog-agent/) repo.
 
+#### Windows Agent v6 installation
+
+Starting with version `>= 6.11`, the Agent v6 must be installed with a chef-datadog
+cookbook version `>= 2.18.0`.
+
+For Agent v6 versions `< 6.11`, previous versions of the datadog cookbook
+could still be used, even if we recommend in most of the case to use the
+latest one.
+
+This is due to the Agent v6 running with an unprivileged user on Windows
+since 6.11. However, prior to chef-datadog 2.18.0, the datadog cookbook was
+enforcing Administrators privileges to the Datadog Agent directories and files.
+
 ### Instructions
 
 1. Add this cookbook to your Chef Server, either by installing with knife or by adding it to your Berksfile:
--- a/README.md
+++ b/README.md
@@ -203,16 +203,12 @@ For general information on the Datadog Agent 6, please refer to the [datadog-age
 
 #### Windows Agent v6 installation
 
-Starting with version `>= 6.11`, the Agent v6 must be installed with a chef-datadog
+Starting with version `>= 6.11`, the Windows Agent v6 must be installed with datadog
 cookbook version `>= 2.18.0`.
 
-For Agent v6 versions `< 6.11`, previous versions of the datadog cookbook
-could still be used, even if we recommend in most of the case to use the
-latest one.
-
 This is due to the Agent v6 running with an unprivileged user on Windows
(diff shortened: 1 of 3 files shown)


olivielpeau:
we need to document that earlier versions of the cookbook won't work with agents v6 >= 6.11

remeh:
I've tested it with versions <6.11 (such as 6.8.0) and it worked well 👍

remeh:
1. Nice catch 👍 
2. I've just tried and it works but the permissions will be the one set by the cookbook instead of the MSI. It means that if someone want to setup 6.11 on Windows for the dd-agent user feature but is using chef-datadog 2.17.0, this user won't have the desired behavior (the permissions will be set to Administrator). I think I'll simply document this specific situation in the README, what do you think? I could also consider to add a warning log by testing `agent6_version` in `2.18.0` but I think it's a bad idea as it's a not-so-good best-effort (won't detect anything when this field is not set).

remeh:
Please note that the CHANGELOG for release 2.18.0 has been updated accordingly: #589

README.md
```diff
@@ -201,6 +201,19 @@ Please review the [attributes/default.rb](https://github.com/DataDog/chef-datado
 
 For general information on the Datadog Agent 6, please refer to the [datadog-agent](https://github.com/DataDog/datadog-agent/) repo.
 
+#### Windows Agent v6 installation
+
+Starting with version `>= 6.11`, the Agent v6 must be installed with a chef-datadog
```
olivielpeau: here I'd mention **Windows** Agent v6 again, for extra clarity

README.md
```diff
@@ -201,6 +201,19 @@ Please review the [attributes/default.rb](https://github.com/DataDog/chef-datado
...
+cookbook version `>= 2.18.0`.
+
+For Agent v6 versions `< 6.11`, previous versions of the datadog cookbook
+could still be used, even if we recommend in most of the case to use the
```
olivielpeau: this paragraph can be removed I think: overall even if it may work in some cases we know the standard use case will be broken, so no need to go into details IMO :)

README.md
olivielpeau: also, there is no `chef-datadog`  cookbook, that's just the name of the repo ;) The cookbook is named `datadog`

README.md
```diff
@@ -201,6 +201,19 @@ Please review the [attributes/default.rb](https://github.com/DataDog/chef-datado
...
+latest one.
+
+This is due to the Agent v6 running with an unprivileged user on Windows
+since 6.11. However, prior to chef-datadog 2.18.0, the datadog cookbook was
```
olivielpeau: same here about `chef-datadog`. You could simply write `prior to v2.18.0` here

providers/monitor.rb
```diff
@@ -10,11 +10,7 @@ def whyrun_supported?
 action :add do
   Chef::Log.debug "Adding monitoring for #{new_resource.name}"
   template ::File.join(yaml_dir, "#{new_resource.name}.yaml") do
-    if node['platform_family'] == 'windows'
-      owner 'Administrators'
-      rights :full_control, 'Administrators'
-      inherits false
-    else
+    unless node['platform_family'] == 'windows'
```
olivielpeau: what about Agent 5, should we also not set permissions here if Agent 5 is installed? :)

remeh: It works this way but you're right that for consistency sake, I'll not change it for Agent 5 (it's unchanged for A5 in `dd-agent.rb`). Nice catch, thanks.

PR #590, opened 2019-03-13
Add a note on the future update to dependencies to remove chef-handler.

remeh:
`chef-handler` has been included in Chef 14 itself. As we'll still support Chef 13 in chef-datadog 3.0.0, we just add a notice about the future removal of the chef-handler dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,11 @@ cookbook 'yum', '~> 3.0'
 
 (in Berkshelf/Librarian format)
 
+Please note that the `chef_handler` dependency **will be removed in chef-datadog
+4.0** because it has been included in Chef 14 itself.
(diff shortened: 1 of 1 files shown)


olivielpeau:
I'm not sure it's worth mentioning this now if it's a change that'll happen in 4.0 🤔 

I'm thinking that, since every breaking change will be carefully documented in the changelog, there's not much value of adding warnings in the readme... Let me know what your thoughts are though.

remeh:
I agree, I did this PR because it was in the todo-list but maybe it was there for 4.0, I'll reformulate and set the milestone to 4.0

remeh:
This is a bit more complicated than that. Closing this PR and we'll address this correctly when the time comes.

PR #595, opened 2019-03-18
Support jmx_custom_jars datadog.conf option

wolf31o2:
As remediation for DataDog support ticket #209172, I was pointed at
DataDog/dd-agent/pull/3648 to set `jmx_custom_jars` in `datadog.conf`,
which is not supported by the current template. Patch the template to
support this option.

Signed-off-by: Chris Gianelloni <wolf31o2@gmail.com>


remeh:
Hey @wolf31o2 once again thank you for your contribution! There is a big chance that it'll be shipped in next release. :)

PR #596, opened 2019-03-22
datadog_monitor doesn't automatically restart the Agent

remeh:
It must now be done explicitly while using the `datadog_monitor` resource. See the change in the README.

This PR has been first created by @someara in https://github.com/DataDog/chef-datadog/pull/323


README.md
```diff
@@ -160,7 +160,7 @@ end
 
 ### Example
 
-This example enables the ElasticSearch integration by using the `datadog_monitor` resource. It provides the instance configuration (in this case: the url to connect to ElasticSearch) and set the `use_integration_template` flag to use the default configuration template.
+This example enables the ElasticSearch integration by using the `datadog_monitor` resource. It provides the instance configuration (in this case: the url to connect to ElasticSearch) and set the `use_integration_template` flag to use the default configuration template. Also, it notifies the `service[datadog-agent]` resource in order to restart the Agent. If you use multiple times the resource `datadog_monitor` in your cookbook, it's recommended for speed purpose to restart the Agent only with the last call.
```
olivielpeau:
```suggestion
This example enables the ElasticSearch integration by using the `datadog_monitor` resource. It provides the instance configuration (in this case: the url to connect to ElasticSearch) and sets the `use_integration_template` flag to use the default configuration template. Also, it notifies the `service[datadog-agent]` resource in order to restart the Agent.
```

(note: no need to use `notifies` only once: by default `notifies` is used with the `:delayed` option, which means that Chef will queue the `notifies` until the end of the Chef run, dedupe them, and actually restart the service only once at the end of the run. Pretty neat stuff)

PR #598, opened 2019-03-25
Notice for Chef < 12.7 users.

remeh:
We will soon drop support for Chef < 12.7, adds a notice for users still using those versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,9 @@ Requirements
 ============
 - chef >= 10.14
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,6 +32,9 @@ Requirements
 ============
 - chef >= 10.14
```
olivielpeau: I think it makes sense to update this to `12.7` as well (so that `master` reflects the support of the latest release)

remeh: Absolutely (this has already been covered in #450 which is the one actually dropping the support for < 12.7)

olivielpeau: oh well, thanks, disregard then :)

PR #599, opened 2019-03-25
Drops Agent v4 compatibility code.

remeh:
Removes the last bit of compatibility code for Agent v4.


recipes/_install-linux.rb
```diff
@@ -54,47 +54,43 @@
...
   # If version specified and lower than 5.x
-  Chef::Log.warn 'Support for Agent pre 5.x will be removed in datadog cookbook version 3.0'
-  # Select correct package name based on attribute
-  dd_pkg_name = node['datadog']['install_base'] ? 'datadog-agent-base' : 'datadog-agent'
+  Chef::Log.error 'Support for Agent pre 5.x has been removed. Please use a 2.x release of the cookbook.'
+  return
+end
+
+# default behavior, remove the `base` package as it is no longer needed
+# TODO(remy): we'll probably want to remove that for release 4.0 of the cookbook
```
olivielpeau: in the general case this would be the right approach, but in the specific case of Agent 4.x I think we can entirely remove this code in the 3.0 release of the cookbook. (I think no Agent v4 is running anywhere anymore)

remeh: Good news! I'll remove that now.

PR #600, opened 2019-03-26
Add datadog_integration resource to install integrations.

remeh:
This PR adds the `datadog_integration` resource to be able to setup Agent integrations directly from a Chef recipe.

### DIFF
--- a/.kitchen.docker.yml
+++ b/.kitchen.docker.yml
@@ -11,7 +11,7 @@ platforms:
   # - name: ubuntu-12.04
   - name: centos-6.6
   - name: centos-7.7
-  - name: debian-7.7
+  - name: debian-8.11
 
 suites:
 - name: dd-agent-handler
--- a/resources/integration.rb
+++ b/resources/integration.rb
@@ -7,7 +7,6 @@
 default_action :install
 
 property :name, String, name_attribute: true
-property :cookbook, String, default: 'datadog'
 
 property :version, String, required: true
 
--- a/resources/integration.rb
+++ b/resources/integration.rb
@@ -1,0 +1,58 @@
+# Install/remove a Datadog integration
+# This resource basically wraps the datadog-agent integration command to be able
+# to install and remove integrations from a Chef recipe.
+# The datadog_resource must be used on a system where the datadog-agent has
+# already been setup.
+
+default_action :install
+
+property :name, String, name_attribute: true
+property :cookbook, String, default: 'datadog'
+
+property :version, String, required: true
+
+action :install do
+  unless node['datadog']['agent6']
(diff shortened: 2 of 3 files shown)


resources/integration.rb
```diff
@@ -0,0 +1,57 @@
...
+      level :error
+    end
+    return
+  end
+
+  log "Getting integration #{new_resource.name}" do
+    level :debug
+  end
+
+  execute 'integration install' do
```
truthbk: might require running with `elevated: true` on windows? right? I'm not sure but be sure to test out.

resources/integration.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+    return
+  end
+
+  log "Removing integration #{new_resource.name}" do
+    level :debug
+  end
+
+  execute 'integration remove' do
+    command   "\"#{agent_exe_filepath}\" integration remove #{new_resource.name}"
```
truthbk: Is this "idempotent", or will calling the `integration remove` with a bad/removed integration name return a non-zero exit code?

.kitchen.docker.yml
```diff
@@ -11,7 +11,7 @@ platforms:
   # - name: ubuntu-12.04
   - name: centos-6.6
   - name: centos-7.7
-  - name: debian-7.7
+  - name: debian-8.11
```
truthbk: why kill `debian-7.7`? Incompatibility?

remeh: It's just for now, CircleCI use the regular wheezy docker images which are not working anymore since Debian has put wheezy in archive. I'll restore wheezy tests when debian/eol images of wheezy will be available.

remeh: See https://lists.debian.org/debian-devel-announce/2019/03/msg00006.html

resources/integration.rb
remeh: Good one! The `integration remove` return a 0 exit code if the integration is not installed so it's fine :+1:

.kitchen.docker.yml
truthbk: Jessie was also removed from the mirror network, and moved into the archive. I guess we'll need to wait for the updated builders and perhaps bump all the way to _stretch_ to test this out.

resources/integration.rb
remeh: Good catch. We can't really use `elevated: true` because it would mean to ask for user credentials to execute this. I've added a note for Chef Windows users that the running chef-client must have read access to the `datadog.yaml`file.

olivielpeau:
A bit late to the party, but this looks great! 👏

PR #602, opened 2019-04-10
Stop uninstalling/reinstalling the package each time on Windows.

remeh:
As the MSI is creating a custom user, the behavior of uninstalling/reinstalling
was creating a new user each time.

### DIFF
--- a/recipes/_install-windows.rb
+++ b/recipes/_install-windows.rb
@@ -85,14 +85,14 @@ remote_file temp_file do
   notifies :remove, 'package[Datadog Agent removal]', :immediately
 end
 
-# Install the package
+# Install/upgrade the package
 windows_package 'Datadog Agent' do # ~FC009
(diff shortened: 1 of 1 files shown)


olivielpeau:
@remeh note: Starting with 6.11.0, the agent MSI won't allow downgrades anymore. Without this change I think there would be no impact on the cookbook (since for a downgrade it would uninstall the Agent and then install the lower version), but with this change I _think_ this  would need to be handled differently...
Could you give the downgrade a try as well? We can then figure out whether we need to do anything specific for downgrades (or just add docs, or do nothing if everything works)

albertvaka:
I can confirm this change breaks the downgrades. We would need a way to use the previous behaviour conditionally if we are downgrading.

AliDatadog:
Closed for inactivity.

PR #606, opened 2019-04-18
Notice on the Windows downgrade unavailable on Windows.

remeh:
We can't downgrade Agent 6 on Windows using chef-datadog for now.


albertvaka:
I can downgrade just fine, since we currently uninstall + install we don't go through the path of a real downgrade. Should we close this?

olivielpeau:
I'm OK with closing this (can't remember exactly what led us to think downgrades didn't work), but such a note may have to be introduced depending on how the conversation in https://github.com/DataDog/chef-datadog/pull/602 goes

AliDatadog:
Closed for inactivity.

PR #607, opened 2019-04-23
Do not set DATADOG_HOST anymore.

remeh:
By using this PR in chef-handler-datadog, we don't need to set DATADOG_HOST anymore.

https://github.com/DataDog/chef-handler-datadog/pull/103/files


remeh:
Already done in #582

PR #615, opened 2019-05-16
Old windows cookbook does not support the `returns` attribute.

remeh:
We advice to use `windows ~> 1.38.0` for some setup, but this cookbook version does not  support the `returns` attribute yet.

### DIFF
--- a/recipes/_install-windows.rb
+++ b/recipes/_install-windows.rb
@@ -82,5 +82,9 @@ windows_package 'Datadog Agent' do # ~FC009
   installer_type installer_type
   options install_options
   action :install
-  returns [0, 3010]
+  if respond_to?(:returns)
+    returns [0, 3010]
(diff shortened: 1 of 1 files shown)


truthbk:
Tests are failing though!

olivielpeau:
Does it still make sense to advise users to use `windows ~> 1.38.0` with the future 3.x versions of the cookbook ? 🤔Shouldn't we not have any dependency on the `windows` cookbook anymore?

remeh:
@olivielpeau You're right we don't have this dep anymore, I've tested on 12/13/14 and it works correctly without the `windows` cookbook.
I'll create a PR to remove the advice in the README + comment this fix.

remeh:
Done in https://github.com/DataDog/chef-datadog/pull/617

PR #617, opened 2019-05-20
We've no dependency to the windows cookbook anymore.

remeh:


recipes/_install-windows.rb
```diff
@@ -82,6 +82,8 @@
   installer_type installer_type
   options install_options
   action :install
+  # Before 3.0.0, we adviced users to use the windows cookbook ~> 1.38.0,
+  # we should probably keep the compatibilty for some time.
```
olivielpeau: is there a reason to remove it and make this a breaking change in 3.0.0?

remeh: @olivielpeau With the newer versions for Chef we support we don't need to pin it anymore and this is not a breaking change.
The reason I want to keep with this `respond_to?` for this major (we should remove it for chef-datadog 4.0) is that we've previously advised users of the cookbook to pin 1.38.0, so many users could still have this windows version pinned.

olivielpeau: ok for removing it in the next major 👍

PR #618, opened 2019-05-20
Allow users to set a custom user/password to run the Windows Datadog Agent service.

remeh:
Let the user of the cookbook provide custom credentials for the Datadog Agent service. It only makes sense for Agents >= `6.11.0` (the feature appeared in the Agent in `6.11.0`).

Note that this PR targets `v2.x` and will later be cherry-picked in `master` for 3.x versions

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -197,14 +197,13 @@ else
 end
 
 # Since 6.11.0, the Datadog Agent creates/uses a custom user to run on Windows.
-# Set `windows_ddagentuser_name` using the format `<domain>/<user>` to provide a
+# Set `windows_ddagentuser_name` using the format `<domain>\<user>` to provide a
 # specific username and `windows_ddagentuser_password` to provide a specific password.
 # You can set those values here, in role/environment/node or set in your
 # node `run_state` under the keys `['datadog']['windows_ddagentuser_name']` and
 # `['datadog']['windows_ddagentuser_password']`.
 default['datadog']['windows_ddagentuser_name'] = nil
 default['datadog']['windows_ddagentuser_password'] = nil
-
 
 # DEPRECATED, will be removed after the release of datadog-agent 6.0
 # Set to true to always install datadog-agent-base (usually only installed on
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -196,6 +196,16 @@ else
   default['datadog']['agent_name'] = 'datadog-agent'
 end
 
+# Since 6.11.0, the Datadog Agent creates/uses a custom user to run on Windows.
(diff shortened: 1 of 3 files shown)


attributes/default.rb
```diff
@@ -196,6 +196,16 @@
   default['datadog']['agent_name'] = 'datadog-agent'
 end
 
+# Since 6.11.0, the Datadog Agent creates/uses a custom user to run on Windows.
+# Set `windows_ddagentuser_name` using the format `<domain>/<user>` to provide a
```
derekwbrown: `<domain>\<user>`

remeh: Whoops. Nice catch!

recipes/_install-windows.rb
```diff
@@ -55,6 +55,12 @@
   installer_type = :msi
   # Agent >= 5.12.0 installs per-machine by default, but specifying ALLUSERS=1 shouldn't affect the install
   install_options = '/norestart ALLUSERS=1'
+
+  # Since 6.11.0, the core and APM/trace components of the Windows Agent run under
+  # a specific user instead of LOCAL_SYSTEM, check whether the user has provided
+  # custom credentials and use them if that's the case.
+  install_options.concat(' DDAGENTUSER_NAME=').concat(Chef::Datadog.ddagentuser_name(node)) if Chef::Datadog.ddagentuser_name(node)
+  install_options.concat(' DDAGENTUSER_PASSWORD=').concat(Chef::Datadog.ddagentuser_password(node)) if Chef::Datadog.ddagentuser_password(node)
```
olivielpeau: is there a chance that there are special characters in the name or password that may need to be escaped/quoted when passed through the command line?

attributes/default.rb
```diff
@@ -196,6 +196,15 @@
   default['datadog']['agent_name'] = 'datadog-agent'
 end
 
+# Since 6.11.0, the Datadog Agent creates/uses a custom user to run on Windows.
+# Set `windows_ddagentuser_name` using the format `<domain>\<user>` to provide a
+# specific username and `windows_ddagentuser_password` to provide a specific password.
+# You can set those values here, in role/environment/node or set in your
+# node `run_state` under the keys `['datadog']['windows_ddagentuser_name']` and
+# `['datadog']['windows_ddagentuser_password']`.
```
olivielpeau: just a nit on this sentence: I'd clarify this to:

```
You can set these 2 values as node attributes or on `node.run_state` under the keys `['datadog']['windows_ddagentuser_name']` and `['datadog']['windows_ddagentuser_password']`
```

recipes/_install-windows.rb
remeh: I've tested a password with some `'"[](){}!-=/\#~.` so I think we're fine.

PR #621, opened 2019-05-23
Do not use the `log` resource to emit some execution log.

remeh:
Using the Chef `log` resource invalidate our custom resources each time they've called, for example restarting the datadog-agent each time the `datadog_monitor` resource is called.


remeh:
You're right. Removed the last call to log. I'll work on `datadog_integration` in another PR. Thanks 👍

PR #623, opened 2019-05-29
Fix integrations tests path.

remeh:
Fix integrations test.

### DIFF
--- a/test/integration/datadog_consul/serverspec/consul_spec.rb
+++ b/test/integration/datadog_consul/serverspec/consul_spec.rb
@@ -1,5 +1,6 @@
 # Encoding: utf-8
 
+require 'spec_helper'
 require 'json_spec'
 require 'serverspec'
 require 'yaml'
@@ -7,9 +8,9 @@ require 'yaml'
 set :backend, :exec
 set :path, '/sbin:/usr/local/sbin:$PATH'
 
-AGENT_CONFIG = '/etc/datadog-agent/conf.d/consul.yaml'.freeze
+AGENT_CONFIG = File.join(@agent_config_dir, 'conf.d/consul.yaml')
 
-describe service('datadog-agent') do
+describe service(@agent_service_name) do
   it { should be_running }
 end
 
--- a/test/integration/datadog_consul/serverspec/consul_spec.rb
+++ b/test/integration/datadog_consul/serverspec/consul_spec.rb
@@ -7,7 +7,7 @@ require 'yaml'
 set :backend, :exec
 set :path, '/sbin:/usr/local/sbin:$PATH'
 
-AGENT_CONFIG = '/etc/dd-agent/conf.d/consul.yaml'.freeze
+AGENT_CONFIG = '/etc/datadog-agent/conf.d/consul.yaml'.freeze
 
 describe service('datadog-agent') do
   it { should be_running }
@@ -29,6 +29,7 @@ describe file(AGENT_CONFIG) do
           'tags' => ['_default']
         }
       ],
+      'logs' => nil,
       'init_config' => nil
     }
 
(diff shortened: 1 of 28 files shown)


test/integration/datadog_consul/serverspec/consul_spec.rb
```diff
@@ -7,7 +7,7 @@
 set :backend, :exec
 set :path, '/sbin:/usr/local/sbin:$PATH'
 
-AGENT_CONFIG = '/etc/dd-agent/conf.d/consul.yaml'.freeze
+AGENT_CONFIG = '/etc/datadog-agent/conf.d/consul.yaml'.freeze
```
truthbk: I know this location is supported, but wouldn't it maybe be better to use the default expected location: `/etc/datadog-agent/conf.d/consul.d/conf.yaml`?

Also, have we entirely dropped A5 support? Would we want to keep testing for backward compatibility?

test/integration/datadog_kafka-consumer/bats/kafka_consumer_config.bats
```diff
@@ -1,11 +1,11 @@
 @test "kafka.yaml exists" {
-  [ -f /etc/dd-agent/conf.d/kafka_consumer.yaml ]
+  [ -f /etc/datadog-agent/conf.d/kafka_consumer.yaml ]
```
truthbk: ditto, wouldn't the expected path here be: `/etc/datadog-agent/conf.d/kafka_consumer.d/conf.yaml`?

We should probably also put this in a constant variable so that we don't need to repeat it in multiple places on the file.

sidenote: kind of surprised we're using bats here!

test/integration/datadog_postfix/serverspec/postfix_spec.rb
```diff
@@ -7,7 +7,7 @@
 set :backend, :exec
 set :path, '/sbin:/usr/local/sbin:$PATH'
 
-AGENT_CONFIG = '/etc/dd-agent/conf.d/postfix.yaml'.freeze
+AGENT_CONFIG = '/etc/datadog-agent/conf.d/postfix.yaml'.freeze
```
truthbk: ditto.

test/integration/datadog_ssh_check/serverspec/ssh_check_spec.rb
```diff
@@ -7,7 +7,7 @@
 set :backend, :exec
 set :path, '/sbin:/usr/local/sbin:$PATH'
 
-AGENT_CONFIG = '/etc/dd-agent/conf.d/ssh_check.yaml'.freeze
+AGENT_CONFIG = '/etc/datadog-agent/conf.d/ssh_check.yaml'.freeze
```
truthbk: ditto.

test/integration/datadog_supervisord/serverspec/supervisord_spec.rb
```diff
@@ -7,7 +7,7 @@
 set :backend, :exec
 set :path, '/sbin:/usr/local/sbin:$PATH'
 
-AGENT_CONFIG = '/etc/dd-agent/conf.d/supervisord.yaml'.freeze
+AGENT_CONFIG = '/etc/datadog-agent/conf.d/supervisord.yaml'.freeze
```
truthbk: ditto.

test/integration/datadog_vault/serverspec/vault_spec.rb
```diff
@@ -7,7 +7,7 @@
 set :backend, :exec
 set :path, '/sbin:/usr/local/sbin:$PATH'
 
-AGENT_CONFIG = '/etc/dd-agent/conf.d/vault.yaml'.freeze
+AGENT_CONFIG = '/etc/datadog-agent/conf.d/vault.yaml'.freeze
```
truthbk: ditto - I'm so annoying.

test/integration/datadog_consul/serverspec/consul_spec.rb
remeh: Those are old kitchen tests manually started while working on the cookbook.
It looks like they're not really up-to-date (e.g. logs value), I don't think we mind preserving backward compatibility here.

test/integration/datadog_kafka-consumer/bats/kafka_consumer_config.bats
remeh: I've used constant wherever it was possible, but I think this bats (discovered `bats` this with this projects...) is out of scope :D

PR #626, opened 2019-06-04
[system-probe] add system-probe agent recipe

shang-wang:
This PR adds system-probe recipe to the cookbook. 

cc: @DataDog/burrito @DataDog/agent-core


attributes/default.rb
```diff
@@ -359,6 +359,14 @@
 default['datadog']['process_agent']['container_interval'] = nil
 default['datadog']['process_agent']['rtcontainer_interval'] = nil
 
+# System probe functionality settings
+default['datadog']['system_probe']['enabled'] = nil
```
sunhay: Should this be false?

attributes/default.rb
```diff
@@ -359,6 +359,14 @@
...
+# sysprobe_socket defines the unix socket location
+default['datadog']['system_probe']['sysprobe_socket'] = nil
```
sunhay: I think you can specify the correct default instead of nil

recipes/system-probe.rb
```diff
@@ -0,0 +1,49 @@
...
+
+# Set the system-probe agent service enable or disable
+agent_enable = node['datadog']['system_probe']['enabled'] ? :enable : :disable
+# Set the correct Agent startup action
+agent_start = node['datadog']['system_probe']['enabled'] ? :start : :stop
+
+#
+# Configures system-probe agent
+#
+if node['datadog']['agent6']
```
sunhay: `if node['datadog']['agent6'] and node['datadog']['system_probe']['enabled']`?

recipes/system-probe.rb
```diff
@@ -0,0 +1,49 @@
...
+    )
+    owner 'dd-agent'
+    group 'root'
+    mode '640'
+    notifies :restart, 'service[datadog-agent-sysprobe]', :delayed unless agent_start == false
+    # since process-agent collects network info through system-probe, enabling system-probe should also restart process-agent
+    notifies :restart, 'service[datadog-agent]', :delayed unless agent_start == false
+  end
+else
+  raise "system-probe agent does not run in agent5 environment"
```
sunhay: Will this cause chef to fail?

sunhay: Why not just wrap everything in the if-block above?

attributes/default.rb
```diff
@@ -359,6 +359,14 @@
 default['datadog']['process_agent']['container_interval'] = nil
 default['datadog']['process_agent']['rtcontainer_interval'] = nil
 
+# System probe functionality settings
+default['datadog']['system_probe']['enabled'] = false
+# sysprobe_socket defines the unix socket location
+default['datadog']['system_probe']['sysprobe_socket'] = '/opt/datadog-agent/run/sysprobe.sock'
+# debug_port is the http port for expvar, it is disabled if not specified
+default['datadog']['system_probe']['debug_port'] = nil
```
sunhay: 0?

templates/default/system_probe.yaml.erb
```diff
@@ -0,0 +1,14 @@
+<%
+## Populate system_probe_config ##
+system_probe_config = {
+  system_probe_config: {
+    enabled: node['datadog']['system_probe']["enabled"],
+    sysprobe_socket: node['datadog']['system_probe']['sysprobe_socket'],
+    debug_port: node['datadog']['system_probe']['debug_port'],
+    bpf_debug: node['datadog']['system_probe']['bpf_debug'],
```
leeavital: can you include `enable_conntrack` here?

attributes/default.rb
shang-wang: nil and 0 are the same, right?

recipes/system-probe.rb
```diff
@@ -0,0 +1,49 @@
...
+end
+
+# Common configuration
+service_provider = nil
+if node['datadog']['agent6'] &&
+   (((node['platform'] == 'amazon' || node['platform_family'] == 'amazon') && node['platform_version'].to_i != 2) ||
```
leeavital: Is there a better way to detect upstart?  This seems  like it could be brittle

attributes/default.rb
shang-wang: Yea I wasn't paying attention to those default values for testing, I added a few.

recipes/system-probe.rb
```diff
@@ -0,0 +1,49 @@
...
+    variables(
+      enabled: node['datadog']['system_probe']['enabled'],
+      sysprobe_socket: node['datadog']['system_probe']['sysprobe_socket'],
+      debug_port: node['datadog']['system_probe']['debug_port'],
+      bpf_debug: node['datadog']['system_probe']['bpf_debug']
+    )
+    owner 'dd-agent'
+    group 'root'
+    mode '640'
+    notifies :restart, 'service[datadog-agent-sysprobe]', :delayed unless agent_start == false
```
leeavital: it looks like the values of agent_start and `:start` and `stop` both of which are truthy.

I think you want `unless agent_start == :stop`

shang-wang: the `agent_start` here is a little weird. It also controls the config `enabled`. You are right that a restart would happen regardless of the value of `agent_start`, but system-probe would quit if `enabled: false`, so it'll stop right after it starts.

shang-wang: I could use another flag to just indicate the `enabled` in chef land, but it's kinda duplicate.

recipes/system-probe.rb
shang-wang: I copied this from main datadog-agent for detection, I'm not sure what's a better way.

recipes/system-probe.rb
shang-wang: Oh wait I don't think what I said is what you've asked......in the definition of `agent_start` at the beginning, that ternary operator takes `:start` as true and `:stop` as false.

attributes/default.rb
sunhay: I'm not sure how ruby interprets them. What gets injected into the config?

recipes/system-probe.rb
sunhay: Can we turn this into a helper function that both recipes can call

shang-wang:
@leeavital datadog-agent recipe doesn't include sysprobe, rather it's sysprobe recipe includes datadog-agent.

attributes/default.rb
shang-wang: It's just going to put `0` in the config file. What I'm saying is port 0 is the same in config as empty. To me using empty makes more sense than `0`.

sunhay: Can we be explicit and leave the default as 0 here? I have no idea where the conversion from nil to zero happens.

recipes/dd-agent.rb
```diff
@@ -137,3 +137,5 @@ def template_vars
 
 # Install integration packages
 include_recipe 'datadog::integrations' unless is_windows
+
+include_recipe 'datadog::system-probe'
```
sunhay: perhaps we can do the equivalent of: `unless system_probe_binary_doesnt_exist`?

sunhay: `is_windows` is another safe thing to include, though that would be covered by the `system_probe_binary_doesnt_exist` value

shang-wang: This is unfinished work, it’s not totally working yet, I’ll finish up tomorrow

recipes/dd-agent.rb
```diff
@@ -135,5 +135,14 @@ def template_vars
   retry_delay 5
 end
 
+# only load system-probe recipe if an agent6 installation comes with it
```
remeh: I don't think we should do it this way.
First I have to ask: when do we want to ship the system-probe? Automatically on every installs of Agent 6 (except on Windows) or could it be on user intent?
With the addition of the system-probe recipe, all users of the `chef-datadog` cookbook will be able to choose to add it in their runlist/cookbooks/etc. to install the system-probe on their nodes.

attributes/default.rb
```diff
@@ -359,6 +359,15 @@
 default['datadog']['process_agent']['container_interval'] = nil
 default['datadog']['process_agent']['rtcontainer_interval'] = nil
 
+# System probe functionality settings
+default['datadog']['system_probe']['enabled'] = false
+# sysprobe_socket defines the unix socket location
+default['datadog']['system_probe']['sysprobe_socket'] = '/opt/datadog-agent/run/sysprobe.sock'
+# debug_port is the http port for expvar, it is disabled if not specified
+default['datadog']['system_probe']['debug_port'] = 0
```
remeh: Is 0 a value understood by the system-probe as `no debug port`? Shouldn't we consider having `nil` here instead?

remeh: Is it an interesting parameter for regular users?

recipes/system-probe.rb
```diff
@@ -0,0 +1,58 @@
...
+    enabled: node['datadog']['system_probe']['enabled'],
+    sysprobe_socket: node['datadog']['system_probe']['sysprobe_socket'],
+    debug_port: node['datadog']['system_probe']['debug_port'],
+    bpf_debug: node['datadog']['system_probe']['bpf_debug'],
+    enable_conntrack: node['datadog']['system_probe']['enable_conntrack']
+  )
+  owner 'dd-agent'
+  group 'root'
+  mode '640'
+  notifies :restart, 'service[datadog-agent-sysprobe]', :delayed unless sysprobe_agent_start == false
```
remeh: The tests on `sysprobe_agent_start` don't look consistent:
 - it is set to `false` in `default/attributes.rb` (boolean)
 - line 21 you set it to either `:start` or `:stop` (Symbol)
 - here, you're comparing it to a boolean

recipes/system-probe.rb
```diff
@@ -0,0 +1,58 @@
...
+  # since process-agent collects network info through system-probe, enabling system-probe should also restart process-agent
+  notifies :restart, 'service[datadog-agent]', :delayed unless sysprobe_agent_start == false
```
remeh: Ditto.

recipes/system-probe.rb
```diff
@@ -0,0 +1,58 @@
...
+   (node['platform'] != 'amazon' && node['platform_family'] == 'rhel' && node['platform_version'].to_i < 7))
+  # use Upstart provider explicitly for Agent 6 on Amazon Linux < 2.0 and RHEL < 7
+  service_provider = Chef::Provider::Service::Upstart
+end
+
+service 'datadog-agent-sysprobe' do
+  action [sysprobe_agent_start]
+  provider service_provider unless service_provider.nil?
+  supports :restart => true, :status => true, :start => true, :stop => true
+  subscribes :restart, "template[#{system_probe_config_file}]", :delayed unless sysprobe_agent_start == false
```
remeh: Ditto.

recipes/dd-agent.rb
```diff
@@ -135,5 +135,14 @@ def template_vars
...
+ruby_block 'include system-probe' do
+  block do
+    if ::File.exist?('/opt/datadog-agent/embedded/bin/system-probe') && node['datadog']['agent6'] && !is_windows
```
remeh: I'm not a huge fan of the binary existence test and not sure that it's actually needed. I would say that it's not really future proof compared to the rest of the recipe.
I think we could just consider adding a note in the README (such as the one there is about the logs feature).

recipes/dd-agent.rb
shang-wang: Good question. We don't want people to do extra hustle to add `system-probe` to their run-list, they don't need to know about this if they don't use it, `dd-agent` recipe is all they need. If they want to use system-probe, just set the `enabled` under `system-probe` to `true` then it should run.

attributes/default.rb
shang-wang: It should be 0 here as default(disabled), the comment might be misleading, I'll change it. If specified with a port it exposes expvars and pprof, the main datadog-agent could collect some metrics from this endpoint.

recipes/dd-agent.rb
shang-wang: I think the `include_recipe` has to be guarded with this, otherwise installing any earlier versions of datadog-agent would break `system-probe` cookbook, because the service `system-probe` doesn't exist. Do you have better suggestions?

recipes/system-probe.rb
shang-wang: I meant to use `node['datadog']['system_probe']['enabled'] ==false`, somehow I put `sysprobe_agent_start` in there......

recipes/dd-agent.rb
shang-wang: I said above based on the fact that dd-agent would still include system-probe recipe.

recipes/dd-agent.rb
remeh: I see 👍

recipes/dd-agent.rb
remeh: OK, if we don't want the user to manually add the recipe in the runlist, that's most likely the best option. We will have to think about this constant if some day the binary is renamed or moved!

recipes/system-probe.rb
```diff
@@ -0,0 +1,58 @@
...
+end
+
+# Common configuration
+service_provider = nil
+if node['datadog']['agent6'] &&
+   (((node['platform'] == 'amazon' || node['platform_family'] == 'amazon') && node['platform_version'].to_i != 2) ||
+    (node['platform'] == 'ubuntu' && node['platform_version'].to_f < 15.04) || # chef <11.14 doesn't use the correct service provider
+   (node['platform'] != 'amazon' && node['platform_family'] == 'rhel' && node['platform_version'].to_i < 7))
+  # use Upstart provider explicitly for Agent 6 on Amazon Linux < 2.0 and RHEL < 7
+  service_provider = Chef::Provider::Service::Upstart
```
sunhay: Shouldn't there be a case where we use systemd?

shang-wang: I think if it's nil it's going to use the default, https://github.com/DataDog/chef-datadog/pull/626/files#diff-a9b4a2e1fcc3717856f3e54e31a25c38R55, and according to chef doc that's the preferable way: 
> Generally, it’s best to let the chef-client choose the provider and this is (by far) the most common approach.

PR #629, opened 2019-06-17
[repository] Add dirmngr when installing from apt repository

KSerrania:
Since debian 9 doesn't package `dirmngr` by default, we have to
add it to be able to install the Agent from our 3rd-party repo.

Update: actually, using the newer versions of the `apt` cookbook is a better solution.

### DIFF
--- a/recipes/repository.rb
+++ b/recipes/repository.rb
@@ -26,6 +26,12 @@ when 'debian'
     action :install
   end
 
+  package 'install-dirmngr' do
+    package_name 'dirmngr'
+    action :install
+    ignore_failure true # Can fail on older distros where the dirmngr package does not exist, but shouldn't prevent install.
+  end
+
   uri = node['datadog']['agent6'] ? node['datadog']['agent6_aptrepo'] : node['datadog']['aptrepo']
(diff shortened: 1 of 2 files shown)


recipes/repository.rb
```diff
@@ -26,6 +26,12 @@
     action :install
   end
 
+  package 'install-dirmngr' do
+    package_name 'dirmngr'
```
remeh: We should probably scope this call to debian using `node['platform_family']` info and maybe even to Debian 9? (not sure about the second part).

I'm wondering if it's a good practice to setup a package on the users' hosts.

olivielpeau: @KSerrania Which subsequent step fails exactly when `dirmngr` isn't installed on Deb 9?

The reason I ask is that I'm inclined to think that this should be handled by the resource that actually fails (so maybe the `apt` cookbook, and/or the core chef resource that currently fails). Maybe newer versions of the `apt` cookbook/Chef handle this gracefully?

If this isn't handled at all on the side of the `apt` cookbook/Chef side, then I guess we have to handle it here. In that case we should document inline why we need to do this, and ideally put the resource call behind a conditional that'd check the Debian/Ubuntu version

KSerrania: Actually, `v7.1` of the `apt` cookbook should do what's needed, so I'm going to revert this change and add a line in the `README` to explain this

PR #634, opened 2019-07-17
[chef][windows] allow custom prefix when specifying agent artifact+ve…

truthbk:
…rsion

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -4,7 +4,7 @@ maintainer_email 'package@datadoghq.com'
 license          'Apache-2.0'
 description      'Installs/Configures datadog components'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version          '3.1.0'
+version          '3.1.1'
 chef_version     '>= 12.7'
 source_url       'https://github.com/DataDog/chef-datadog'
 issues_url       'https://github.com/DataDog/chef-datadog/issues'
--- a/metadata.rb
+++ b/metadata.rb
@@ -4,7 +4,7 @@ maintainer_email 'package@datadoghq.com'
 license          'Apache-2.0'
 description      'Installs/Configures datadog components'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version          '3.1.1'
+version          '3.1.0'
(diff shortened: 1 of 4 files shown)


spec/dd-agent_spec.rb
```diff
@@ -268,11 +268,12 @@ def set_env_var(name, value)
           :file_cache_path => 'C:/chef/cache'
         ) do |node|
           node.normal['datadog'] = {
-            'agent6' => false,
+            'agent6' => true,
             'api_key' => 'somethingnotnil',
             'agent6_version' => {
               'debian' => '1:6.9.0-1',
               'rhel' => '6.9.0-1',
+              'windows' => '6.9.0'
```
olivielpeau: nit, you can ignore: trailing commas are nice :)

 ```suggestion
               'windows' => '6.9.0',
```

File: metadata.rb [hunk 2]
```diff
@@ -4,7 +4,7 @@
 license          'Apache-2.0'
 description      'Installs/Configures datadog components'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version          '3.1.0'
+version          '3.1.1'
```
olivielpeau: version bump should be removed from PR

recipes/_install-windows.rb
```diff
@@ -40,8 +40,11 @@
   dd_agent_latest = 'ddagent-cli-latest'
 end
 
+# If a different file prefix is specified, use that:
+dd_agent_installer_prefix = node['datadog']['windows_installer_prefix'] ? node['datadog']['windows_installer_prefix'] : 'ddagent-cli'
```
olivielpeau: The attribute `windows_installer_prefix` must be added to `attributes/default.rb`, and documented there.

Also, nit: for consistency with other Windows-related attributes, I slightly prefer a name such as `windows_agent_installer_prefix`, even though it's longer

metadata.rb
truthbk: Yeah, we're using this from the QA env, so I needed the bump. But we'll remove that before merging.

remeh: Let's remove this and merge to `master`!

PR #638, opened 2019-07-31
[repository] Trust new GPG key only when installing Agent 6 RPM

KSerrania:
In the near future, we're going to sign our RPM package with a
newer GPG key.
To make sure chef installs of Agent 6 work, the new key
should be installed if Agent 6 is installed.
However, since some OSes supported by Agent 5 (but not Agent 6)
are not able to import the new key, we shouldn't import it if Agent 5 is installed.

### DIFF
--- a/recipes/repository.rb
+++ b/recipes/repository.rb
@@ -48,7 +48,7 @@ when 'debian'
   end
 when 'rhel', 'fedora', 'amazon'
   # Import new RPM key if installing Agent 6
-  if node['datadog']['agent6']
+  if node['datadog']['yumrepo_gpgkey_new'] && node['datadog']['agent6']
     # gnupg is required to check the downloaded key's fingerprint
     package 'gnupg' do
       action :install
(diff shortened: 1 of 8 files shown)


recipes/repository.rb
```diff
@@ -47,8 +47,8 @@
     action :remove
   end
 when 'rhel', 'fedora', 'amazon'
-  # Import new RPM key
-  if node['datadog']['yumrepo_gpgkey_new']
+  # Import new RPM key if installing Agent 6
+  if node['datadog']['agent6']
```
remeh: I think we should keep the test on `node['datadog']['yumrepo_gpgkey_new']` just because we also use the value in `node['datadog']['yumrepo_gpgkey_new']` to download the key, thus, we don't want it to be null or empty.

recipes/repository.rb
```diff
@@ -88,8 +88,8 @@
     action :create
   end
 when 'suse'
-  # Import new RPM key
-  if node['datadog']['yumrepo_gpgkey_new']
+  # Import new RPM key if installing Agent 6
+  if node['datadog']['agent6']
```
remeh: Ditto.

olivielpeau:
curious: does the import of the new key actually fail on CentOS 5? (i.e. the version of CentOS that the Agent 5 supports but Agent 6 doesn't support)

Asking because, if so, that means the current code importing the new key has always been failing on CentOS 5.

AliDatadog:
Closed for inactivity.

PR #639, opened 2019-09-03
Update README.md

nicholas-devlin:
wasn't able to figure out how to set nested attributes using the `extra_configs` field, took like an hour or 2 of asking around/ guess and checking, so I added an example

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,15 +263,29 @@ Administrators privileges to the Datadog Agent directories and files.
 Should you wish to add additional elements to the Agent v6 configuration file
 (typically `datadog.yaml`) that are not directly available
 as attributes of the cookbook, you may use the `node['datadog']['extra_config']`
-attribute. 
+attribute. This attribute is a hash and will be marshaled into the configuration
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -263,19 +263,35 @@ Administrators privileges to the Datadog Agent directories and files.
...
+
+E.g.
+
+```ruby
+default['datadog']['extra_config']['secret_backend_command'] = '/sbin/local-secrets'`
+```
+
+For nested attributes, use object syntax: 
+
+E.g.
```
remeh: Good addition, I didn't know it was a problem for nested attributes and that we must use the object syntax 👍 
However, what do you think of sticking with the original example and adding the nested attribute example below? I think it would be more natural to read.

nicholas-devlin: Yeah if you think reordering the addition makes sense, that's totally fine. I'm not entirely sure how you want to rearrange it though. If you want to just make another commit with your recommendation that is probably easiest.

PR #641, opened 2019-09-23
Support Agent package installation on RHEL 8 / Fedora 28+

KSerrania:
### What does this PR do?

Support Agent package installation on RHEL 8 / Fedora 28+.

### Motivation

`yum_package` and `dnf_package` are broken on RHEL 8 ([Chef issue](https://github.com/chef/chef/issues/7988)): `package` resources fail on RHEL 8 due to python not being bundled the same way.

This was [fixed](https://github.com/chef/chef/pull/8003) for the `dnf_resource`, but not for the `yum_resource` ([a PR was merged](https://github.com/chef/chef/pull/8005) but doesn't fix the problem). Thus, `dnf_package` needs to be used.

The minimal Chef version which supports RHEL 8 is [`v15.0.293`](https://github.com/chef/chef/blob/master/CHANGELOG.md#v150293-2019-05-14) (which is the first public Chef 15 version), so this should be added as a requirement if users want to use the cookbook on RHEL 8

### Additional Notes

I didn't confirm that Fedora 28+ is affected (no kitchen tests on Fedora), but:
- it probably is, as Fedora 28 is the version which corresponds to RHEL 8,
- using `dnf` on Fedora is not a problem (it's the default package manager on this platform)..

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Platforms
 * Amazon Linux
 * CentOS
 * Debian
-* RedHat
+* RedHat (RHEL 8 requires chef >= 15)
 * Scientific Linux
 * Ubuntu
 * Windows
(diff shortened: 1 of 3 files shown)


albertvaka:
lgtm

albertvaka:
There are some code style issues that fail the tests though.

PR #644, opened 2019-10-04
test-kitchen >= 1.21.0 prefers kitchen.yml instead of .kitchen.yml

albertvaka:
Also I find it less confusing that this file is not hidden


albertvaka:
I think this should do the trick. Any way to test it before merging?

PR #646, opened 2019-10-08
Agent 7 support

albertvaka:
- Removes `agent6_` variables from the "public API" of the cookbook.
- Adds `agent_major_version`, which can be 5, 6 or 7.
- Updates README and adds upgrade notes.
- Updates tests and adds one test.
- Installs only the new RPM key when installing Agent 7 and adds one test for that as well.


recipes/repository.rb
```diff
@@ -71,15 +79,20 @@
     end
   end
 
+  case node['datadog']['agent_major_version'].to_i
```
remeh: This is broken on CentOS (the package is not found in the repos), I think that's because the `baseurl` always need to be computed but this is not working if `agent_major_version` is not set.
Also, it should have to take care of an inconsistent `agent_version` with a major different that `agent_major_version`.

recipes/repository.rb
```diff
@@ -122,15 +135,20 @@
     action :nothing
   end
 
+  case node['datadog']['agent_major_version'].to_i
```
remeh: Ditto.

metadata.rb
```diff
@@ -4,7 +4,7 @@
 license          'Apache-2.0'
 description      'Installs/Configures datadog components'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version          '3.3.0'
+version          '3.99.0'
```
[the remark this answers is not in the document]
albertvaka: I was thinking about 3.99.1 for fixes, and keep 3.4.0 available in case we want to actually release another 3.x version. I don't think this is important as long as we don't actually release a 3.99 to the public.

attributes/default.rb
```diff
@@ -156,17 +126,21 @@
...
   end
 
+# repos where datadog-agent packages are available
+default['datadog']['aptrepo'] = 'http://apt.datadoghq.com'
+default['datadog']['aptrepo_dist'] = 'stable'
+default['datadog']['yumrepo'] = 'https://yum.datadoghq.com/stable/'
+default['datadog']['yumrepo_suse'] = 'https://yum.datadoghq.com/suse/stable/'
+default['datadog']['agent5_yumrepo'] = "#{yum_protocol}://yum.datadoghq.com/rpm/#{architecture_map[node['kernel']['machine']]}/"
+default['datadog']['agent5_yumrepo_suse'] = "https://yum.datadoghq.com/suse/rpm/#{architecture_map[node['kernel']['machine']]}/"
+
```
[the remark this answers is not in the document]
albertvaka: It's defined later in the file. Do you mean I should move it here?

recipes/repository.rb
albertvaka: `agent_major_version` should always be set.

README.md
```diff
@@ -239,11 +239,13 @@ By default, this cookbook installs Agent v6, if you want to install Agent v5, pl
 
 Attributes are available to have finer control over how you install Agent v6:
 
- * `agent6_version`: allows you to pin the agent version (recommended).
- * `agent6_package_action`: defaults to `'install'`, may be set to `'upgrade'` to automatically upgrade to latest (not recommended, we recommend pinning to a version with `agent6_version` and change that version to upgrade).
- * `agent6_aptrepo`: desired APT repo for the agent. Defaults to `http://apt.datadoghq.com`
- * `agent6_aptrepo_dist`: desired distribution for the APT repo. Defaults to `stable`
- * `agent6_yumrepo`: desired YUM repo for the agent. Defaults to `https://yum.datadoghq.com/stable/6/x86_64/`
+ * `agent_major_version`: allows you to pin the major release of the agent (recommended)
+ * `agent_version`: allows you to pin a specific agent version (recommended).
```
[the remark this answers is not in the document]
albertvaka: The problem with the pinning of minor versions is that you have to specify the version as the package manager wants it (so, its different in deb/rpm/windows). I've tried to keep the changes to the minimum, so I only changed the boolean to a an int, while keeping the `agent_version` field as it was. I agree it's a bit confusing, but it's not worse than what we had.

Having a single version field doesn't let you pin the major version without pinning the minor and patch versions, which I think can be a use case. Maybe we could support stuff like "6.x", but I'm not sure if that's common practice.

recipes/dd-agent.rb
```diff
@@ -44,13 +48,11 @@
 # To add integration-specific configurations, add 'datadog::config_name' to
 # the node's run_list and set the relevant attributes
 #
-if node['datadog']['agent6']
+if node['datadog']['agent_major_version'].to_i > 5
```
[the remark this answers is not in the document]
albertvaka: Should I do this in `recipe_helpers`?

attributes/default.rb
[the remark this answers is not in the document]
remeh: Dismiss, I've missed it :)

recipes/repository.rb
remeh: You're right it's properly set in the default attributes, the problem must be elsewhere then 🤔

metadata.rb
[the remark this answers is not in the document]
remeh: Definitely made this review too early today. `3.99.1` will work. 🙇

README.md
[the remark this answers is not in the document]
remeh: > I agree it's a bit confusing, but it's not worse than what we had.

Fair point, and again, it works well 👍 . My thoughts were about the parts of code where we have the conditions on `agent_major_version` to later use `agent_version` to do something (mainly in `repository.rb`). The value of these fields will be one of the first thing to check in case of errors.

recipes/dd-agent.rb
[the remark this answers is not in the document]
remeh: This could be done the same way it has been done for `is_windows`. Please dismiss if that's too much boilerplate.

recipes/repository.rb
albertvaka: Fixed.

recipes/repository.rb
```diff
@@ -26,17 +26,25 @@
     action :install
   end
 
-  uri = node['datadog']['agent6'] ? node['datadog']['agent6_aptrepo'] : node['datadog']['aptrepo']
-  distribution = node['datadog']['agent6'] ? node['datadog']['agent6_aptrepo_dist'] : node['datadog']['aptrepo_dist']
-  components = node['datadog']['agent6'] ? ['main', '6'] : ['main']
+  case node['datadog']['agent_major_version'].to_i
+  when 7
+    components = ['main', '7']
```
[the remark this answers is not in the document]
albertvaka: I would rather do `components = ['7']`, unless there is a good reason to keep old versions.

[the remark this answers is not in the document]
olivielpeau: Let's clean this up and only have `components = ['7']` 👍 

If I recall correctly the historical reason we have `['main', '6']` for Agent 6 is that early beta versions of Agent 6 were released on the `main` component (at the time we didn't have a `6` component yet). This is not relevant anymore and should be cleaned up.

recipes/dd-agent.rb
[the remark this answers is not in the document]
albertvaka: Most of the checks are `version > 5`, so then that would become `!is_agent5`, which is a bit confusing. I like the current way better.

README.md
```diff
@@ -410,11 +273,172 @@ AWS OpsWorks Chef Deployment
 
 1. Add Chef Custom JSON:
   ```json
-  {"datadog":{"agent6": true, "api_key": "<API_KEY>", "application_key": "<APP_KEY>"}}
+  {"datadog":{"agent_major_version": 6, "api_key": "<API_KEY>", "application_key": "<APP_KEY>"}}
```
KSerrania:
```suggestion
  {"datadog":{"agent_major_version": 7, "api_key": "<API_KEY>", "application_key": "<APP_KEY>"}}
```

recipes/repository.rb
```diff
@@ -26,17 +26,25 @@
...
   end
 
-  uri = node['datadog']['agent6'] ? node['datadog']['agent6_aptrepo'] : node['datadog']['aptrepo']
-  distribution = node['datadog']['agent6'] ? node['datadog']['agent6_aptrepo_dist'] : node['datadog']['aptrepo_dist']
-  components = node['datadog']['agent6'] ? ['main', '6'] : ['main']
+  case node['datadog']['agent_major_version'].to_i
+  when 7
+    components = ['7']
+  when 6
+    components = ['main', '6']
```
KSerrania: Do we really need `main` when installing Agent 6? I thought it only contained Agent 5 and lower packages.
For example, the ansible module only uses the `6` component to install Agent 6: https://github.com/DataDog/ansible-datadog/blob/dbe9a8fa0e03831285146483404b4520670a2174/defaults/main.yml#L24

resources/monitor.rb
```diff
@@ -53,9 +53,7 @@
...
 
 def yaml_dir
-  if node['datadog']['agent6']
-    ::File.join(node['datadog']['agent6_config_dir'], 'conf.d')
-  else
-    ::File.join(node['datadog']['config_dir'], 'conf.d')
-  end
+  is_agent5 = node['datadog']['agent_major_version'].to_i > 5
+  is_windows = node['platform_family'] == 'windows'
+  return is_windows ? "#{ENV['ProgramData']}/Datadog/conf.d" : (is_agent5 ? '/etc/dd-agent/conf.d' : '/etc/datadog-agent/conf.d')
```
KSerrania: `return` is not needed, and RuboCop doesn't like nested ternary operators
```suggestion
 if is_windows
   "#{ENV['ProgramData']}/Datadog/conf.d"
 else 
   is_agent5 ? '/etc/dd-agent/conf.d' : '/etc/datadog-agent/conf.d'
 end
```

recipes/repository.rb
albertvaka: Apparently old versions of Agent 6 were deployed to main. We should check if all the versions are in `6` before removing it. I'll add a comment.

attributes/default.rb
```diff
@@ -193,18 +167,6 @@
...
-# Values that differ on Windows
-# The location of the config folder (containing conf.d)
-# The name of the dd agent service
-# The log file directory (see logging section below)
-if node['platform_family'] == 'windows'
-  default['datadog']['config_dir'] = "#{ENV['ProgramData']}/Datadog"
-  default['datadog']['agent_name'] = 'DatadogAgent'
-else
-  default['datadog']['config_dir'] = '/etc/dd-agent'
-  default['datadog']['agent_name'] = 'datadog-agent'
```
olivielpeau: Ideally, for attributes that aren't defined and shouldn't be used anymore in 4.x, you could add logic to print warnings if they're set.

Note that `config_dir` is now configurable on the Windows Agent since v6.11 (with `APPLICATIONDATADIRECTORY`, see https://docs.datadoghq.com/agent/basic_agent_usage/windows/?tab=agentv6). Doesn't mean that you should add support for this now (3.x didn't support the option OOTB), but, just so you know, we may want to re-introduce this attribute at some point :)

recipes/repository.rb
```diff
@@ -71,14 +79,19 @@
     end
   end
 
+  case node['datadog']['agent_major_version'].to_i
+  when 6, 7
+    baseurl = URI.join(node['datadog']['yumrepo'], "#{node['datadog']['agent_major_version']}/", 'x86_64/').to_s
```
olivielpeau: we could start using `node['kernel']['machine']` here instead of hardcoding `x86_64` (if that helps with ARM support on RHEL-like platforms, should be checked).

Not necessary for this PR, just wanted to point it out.

recipes/repository.rb
```diff
@@ -26,17 +26,25 @@
...
   end
 
-  uri = node['datadog']['agent6'] ? node['datadog']['agent6_aptrepo'] : node['datadog']['aptrepo']
-  distribution = node['datadog']['agent6'] ? node['datadog']['agent6_aptrepo_dist'] : node['datadog']['aptrepo_dist']
-  components = node['datadog']['agent6'] ? ['main', '6'] : ['main']
+  case node['datadog']['agent_major_version'].to_i
+  when 7
+    components = ['7']
+  when 6
+    components = ['main', '6'] # early versions of A6 were deployed to main. TODO: remove main if 6 contains all versions
```
olivielpeau: Checked this: we started releasing Agent 6 to the `6` component starting with `6.0.0-beta.9`. Earlier betas were only released on the `main` component. See:

* [Packages file on the `beta` distribution of `6`](https://s3.amazonaws.com/apt.datadoghq.com/dists/beta/6/binary-x86_64/Packages)
* [Packages file on the `beta` distribution of `main`](https://s3.amazonaws.com/apt.datadoghq.com/dists/beta/main/binary-x86_64/Packages)

At the moment almost nobody uses 6.0.0 betas that are earlier than `beta.9`, so I think it's safe to drop `main` here. Ping me on chat if you want more precise usage stats on early 6.0.0 betas :)

README.md
```diff
@@ -69,6 +71,217 @@ dd-handler recipe. The known workaround is to update your dependency to `chef_ha
...
+|---|---|---|
+| Install Agent 7.x | Not supported |  `'agent_major_version' => 7` |
+| Install Agent 6.x | `'agent6' => true`  |  `'agent_major_version' => 6` |
+| Install Agent 5.x | `'agent6' => false`  |  `'agent_major_version' => 5` |
+| Pin agent version | `'agent_version'` or `'agent6_version'`  |  `'agent_version'` for all versions |
+| Change package_action | `'agent_package_action'` or `'agent6_package_action'`  |  `'agent_package_action'` for all versions |
+| Change APT repo | `'aptrepo'` or `'agent6_aptrepo'`  |  `'aptrepo'` for all versions |
+| Change APT dist | `'aptrepo_dist'` or `'agent6_aptrepo_dist'`  |  `'aptrepo_dist'` for all versions |
+| Change YUM repo | `'yumrepo'` or `'agent6_yumrepo'`  |  `'yumrepo'` for all versions |
+| Change SUSE repo | `'yumrepo_suse'` or `'agent6_yumrepo_suse'`  |  `'yumrepo_suse'` for all versions |
```
olivielpeau: this isn't strictly true: `agent5_yumrepo` and `agent5_yumrepo_suse` exist in 4.x as well, and could be mentioned here

README.md
```diff
@@ -69,6 +71,217 @@ dd-handler recipe. The known workaround is to update your dependency to `chef_ha
...
+  )
+```
+
+The same works for version 5.
+
+### Instructions
+
+1. Add this cookbook to your Chef Server, either by installing with knife or by adding it to your Berksfile:
+  ```
+  cookbook 'datadog', '~> 4.0.0'
```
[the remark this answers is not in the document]
albertvaka: We had 3.0.0 before, so I left what we had.

README.md
```diff
@@ -69,6 +71,217 @@ dd-handler recipe. The known workaround is to update your dependency to `chef_ha
...
+      'agent_major_version' => 7,
+      'agent_version' => '1:7.15.0-1', # optional but recommended
+      'agent_package_action' => 'install',
+    }
+  )
+```
+
+The same applies if upgrading from 5 to 6.
+
+Note that there are Agent v6 counterparts to several well known Agent v5 attributes (code [here](https://github.com/DataDog/chef-datadog/blob/master/attributes/default.rb#L31-L75))
```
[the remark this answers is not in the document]
albertvaka: I've removed that line instead.

recipes/repository.rb
albertvaka: Olivier checked and it can be removed 👌

README.md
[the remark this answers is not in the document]
olivielpeau: ok, no need to change it now yup, we can consider this later

README.md
```diff
@@ -171,22 +173,23 @@ The same works for version 5.
 
 #### Upgrade from Agent v6 to Agent v7
 
-To upgrade from an already installed Agent v6 to Agent v7, you'll have to set the `agent_package_action` to `install` and we recommend to pin to a specific version:
+To upgrade from an already installed Agent v6 to Agent v7, you'll have to either:
+
+* Set `agent_major_version` to `7`, `agent_package_action` to `upgrade` and pin a specific v7 version as `agent_version` (recommended).
```
olivielpeau: note: as long as `agent_version` is set to a version, `agent_package_action` does not have to be set to `upgrade` (`install` is fine)

albertvaka: Copy-paste error!

README.md
```diff
@@ -69,6 +71,220 @@ dd-handler recipe. The known workaround is to update your dependency to `chef_ha
...
+| Action | Cookbook 3.x  | Cookbook 4.x  |
+|---|---|---|
+| Install Agent 7.x | Not supported |  `'agent_major_version' => 7` |
+| Install Agent 6.x | `'agent6' => true`  |  `'agent_major_version' => 6` |
+| Install Agent 5.x | `'agent6' => false`  |  `'agent_major_version' => 5` |
+| Pin agent version | `'agent_version'` or `'agent6_version'`  |  `'agent_version'` for all versions |
+| Change package_action | `'agent_package_action'` or `'agent6_package_action'`  |  `'agent_package_action'` for all versions |
+| Change APT repo | `'aptrepo'` or `'agent6_aptrepo'`  |  `'aptrepo'` for all versions |
+| Change APT dist | `'aptrepo_dist'` or `'agent6_aptrepo_dist'`  |  `'aptrepo_dist'` for all versions |
+| Change YUM repo | `'yumrepo'` or `'agent6_yumrepo'`  |  `'yumrepo'` for Agent 6/7 or `'agent5_yumrepo'` for Agent 5 |
```
julien-lebot: For Agent 6/7 I think we should mention that the base URL changes as well:

```
 'yumrepo' => 'https://yum.datad0g.com/stable/6/x86_64/'
```
becomes
```
 'yumrepo' => 'https://yum.datad0g.com/stable/',
```
Same for `yumrepo_suse`

albertvaka: Added, although it breaks the layout a bit :/
https://github.com/DataDog/chef-datadog/blob/albertvaka/4.0-refactor/README.md

julien-lebot: Yeah but it looks legible to me so I think that's fine 👍

CHANGELOG.md
```diff
@@ -1,6 +1,19 @@
 Changes
 =======
 
+# 4.0.0 / 2019-12-19
+
+## Breaking changes
+
+  * **This cookbook will install Agent 7.x by default**. Datadog Agent 7 uses Python 3 by
+  default so if you were running any custom checks written in Python, they must now be
```
KSerrania: I'd remove the "by default", Agent 7 has no way to run Python 2 even with non-default config.

albertvaka: 👍

PR #648, opened 2019-10-18
Add a setting to disable writing system-probe.yaml

albertvaka:


attributes/default.rb
```diff
@@ -334,6 +334,10 @@
 default['datadog']['process_agent']['rtcontainer_interval'] = nil
 
 # System probe functionality settings
+
+# Wheter this cookbook should write system-probe.yaml or not.
```
KSerrania:
```suggestion
# Whether this cookbook should write system-probe.yaml or not.
```

PR #651, opened 2019-11-07
[windows] safely handle buggy versions

truthbk:


recipes/_install-windows.rb
```diff
@@ -85,14 +85,42 @@
...
 remote_file temp_file do
   source node['datadog']['windows_agent_url'] + dd_agent_installer
   checksum node['datadog']['windows_agent_checksum'] if node['datadog']['windows_agent_checksum']
   retries package_retries unless package_retries.nil?
   retry_delay package_retry_delay unless package_retry_delay.nil?
+  verify do |path|
+    require 'digest'
+
+    unsafe = unsafe_hashsums.include? Digest::SHA256.file(path).hexdigest
+    Chef::Log.info("\n#{fix_message}\n") if unsafe
```
olivielpeau: should probably be at the `error` level since we'll fail the Chef run on this

recipes/_install-windows.rb
```diff
@@ -85,14 +85,42 @@
...
+
+    # verify will abort update if false
+    !unsafe
+  end if not node['datadog']['windows_blacklist_silent_fail']
+
+  # validate the downloaded MSI is safe
```
olivielpeau: nit: I think this comment is meant to be at the top of the `verify` block?

recipes/_install-windows.rb
```diff
@@ -111,4 +139,12 @@
   else
     success_codes [0, 3010]
   end
+  not_if do
+    require 'digest'
+
+    unsafe = unsafe_hashsums.include? Digest::SHA256.file(temp_file).hexdigest
+    Chef::Log.info("\n#{fix_message}\n") if unsafe
```
olivielpeau: nit: we are not failing the chef run on this, but we are silently not installing the Agent. `warning` could be abetter log level here. Also, an additional line about the fact that we're `continuing without installing Datadog Agent` would be helpful :)

Also, note that the recipes that run after this recipe may rely on the Agent to be installed to work, so on hosts that do not already have the Agent installed subsequent recipes may fail the Chef run. Not avoidable IMO, but it may be useful to be document this on the `windows_blacklist_silent_fail` attribute? Up to you, I don't have a strong opinion on this.

truthbk: indeed, the chef runs might fail after depending on circumstances. Not sure we can do much more than what we're currently doing. The attribute is a best effort approach. Will add a comment to it.

README.md
```diff
@@ -65,6 +65,35 @@ cookbook 'yum', '~> 3.0'
 
 (in Berkshelf/Librarian format)
 
+Known Issues
+============
+
+Due to a critical bug in agent versions `6.14.0` and `6.14.1` for Windows, these versions have
+been blacklisted in cookbook versions `>=3.4.0`. `>=2.20.0`.
+
+*PLEASE NOTE:* chef-runs will fail on windows when updating to cookbook versions `>=3.4.0`. `>=2.20.0`
```
truthbk:
```suggestion
**PLEASE NOTE:** chef-runs will fail on windows when updating to cookbook versions `>=3.4.0`. `>=2.20.0`
```

README.md
```diff
@@ -65,6 +65,35 @@ cookbook 'yum', '~> 3.0'
...
+
+If you are updating from **6.14.0 or 6.14.1 on Windows**, we **strongly** recommend following these steps:
+
+1. Install the latest datadog cookbook on your Chef Server (`>=3.4.0` or `>=2.20.0`)
+2. Remove ALL prior versions of the datadog cookbook to ensure only the latest version of the cookbook
+is available to your hosts. Please make sure to backup any modified or forked versions of the datadog
+cookbook using `knife` or `berks` if you need to reference them later.
+3. Set the `agent6_version` attribute to `6.14.2`
+
+*PLEASE NOTE:* It is important to ensure that your hosts ONLY use the latest version of the
```
truthbk:
```suggestion
**PLEASE NOTE:** It is important to ensure that your hosts ONLY use the latest version of the 
```

recipes/_install-windows.rb
```diff
@@ -85,14 +85,42 @@
...
+    Chef::Log.error("\n#{fix_message}\n") if unsafe
+
+    # verify will abort update if false
+    !unsafe
+  end unless node['datadog']['windows_blacklist_silent_fail']
+
+  notifies :run, 'powershell_script[datadog_6.14.x_fix]', :immediately
+end
+
+powershell_script 'datadog_6.14.x_fix' do
```
julien-lebot: Instead of downloading the file manually, we can use Chef's `remote_file` feature.

Something like:

```
fix_ps1 = File.join(Chef::Config['file_cache_path'], 'fix_6_14.ps1')

remote_file fix_ps1 do
    action :create
    backup false
    source 'https://s3.amazonaws.com/ddagent-windows-stable/scripts/fix_6_14.ps1'
end

powershell_script 'Patch system' do
    cwd Chef::Config['file_cache_path']
    command fix_ps1
end
```

PR #654, opened 2019-11-12
[windows] add MSI max timeout knob

truthbk:
MSI operations like install/uninstall might timeout on small cloud instances. Adding this knob to allow customers to work around the issue.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -193,7 +193,7 @@ default['datadog']['windows_blacklist_silent_fail'] = false
 
 # Attribute to specify timeout in seconds on MSI operations (install/uninstall)
 # Default should suffice, but provides a knob in case instances with limited resources timeout.
-default['datadog']['windows_msi_timeout'] = 600
+default['datadog']['windows_msi_timeout'] = 900
 
 # Agent installer checksum
 # Expected checksum to validate correct agent installer is downloaded (Windows only)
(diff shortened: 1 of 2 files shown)


attributes/default.rb
```diff
@@ -191,6 +191,10 @@
 # the blacklist will be resolved enabling this feature.
 default['datadog']['windows_blacklist_silent_fail'] = false
 
+# Attribute to specify timeout on MSI operations (install/uninstall)
+# Default should suffice, but provides a knob in case instances with limited resources time out.
+default['datadog']['windows_msi_timeout'] = 600
```
remeh: Are we sure this default value is enough? The MSI is capable of being veeery slow.

olivielpeau: unit is `seconds`, let's mention that in the comment.

`600` appears to be the default of the windows package resource (https://docs.chef.io/resource_windows_package.html), if we know our MSI can take a bit more than that in some cases we may want to increase our default a bit indeed.

truthbk: The default should be fine most of the time, otherwise we would've heard about this sooner. It's the sort of thing customers will typically encounter with very small instances, I think the default should be fine. 

And absolutely let's specify the units are seconds.

olivielpeau: discussed this offline, let's use `900` seconds as that's the timeout that's used for azure extensions

olivielpeau: @truthbk note: apparently Chef kills the process when the timeout is reached, see the underlying library code: https://github.com/chef/mixlib-shellout/blob/c35d241209d5067afd6ed5fd23c063bf1ae74a58/lib/mixlib/shellout/windows.rb#L120-L135

So, probably another reason to increase this timeout by default. (if the process wasn't killed, we could have relied on the process to complete before the next chef run, but since it is killed, there's a big chance that it would happen on every chef run until the timeout is increased)

PR #662, opened 2019-11-18
Add a Gemfile

albertvaka:
Pin a version of chef that works with the (5-years-unmaintained) emeril gem we use to run `rake release`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,17 @@
+source 'https://rubygems.org'
+
+# Gems here are used to run `rake release`.
+
+group :release do
+  gem 'emeril'
+  gem 'rake'
+  gem 'chef', '~> 12.22'
+  gem 'foodcritic'
+  gem 'chefspec'
+  gem 'cookstyle'
+  gem 'coveralls'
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -0,0 +1,13 @@
+source "https://rubygems.org"
```
KSerrania: rubocop doesn't like double-quoted strings when they are not used for interpolation / special symbols.

```suggestion
source 'https://rubygems.org'
```

Gemfile
```diff
@@ -0,0 +1,13 @@
...
+
+gem 'emeril'
```
KSerrania: Since all this gem is only necessary to release the cookbook (and not to use it), should we put it (and all other `rake release` dependencies) in a group? For example:
```ruby
group :release do
  gem 'emeril'
  ...
end
```
This makes it clear that they're not needed to use the cookbook.

albertvaka:
Pinning chef 12 means that test on local fail due to changes in Chef, so I don't think we can merge this as it is.

PR #663, opened 2019-11-28
Update vagrant boxes used for kitchen tests

albertvaka:
The old ones don't exist anymore, they got renamed


remeh:
Because you're changing the source for the images, you'll have to change the box you want to use: for instance, `debian-7.7` is not available there but `debian-7.9` and `debian-7.10` are, etc.

PR #665, opened 2019-12-11
Autodetect agent_major_version from agent_version

albertvaka:


recipes/dd-agent.rb
```diff
@@ -54,7 +54,7 @@
 # To add integration-specific configurations, add 'datadog::config_name' to
 # the node's run_list and set the relevant attributes
 #
-if node['datadog']['agent_major_version'].to_i > 5
+if Chef::Datadog.agent_major_version(node) > 5
```
KSerrania: `Chef::Datadog.agent_major_version(node)` is used multiple times in this recipe, and each time it is used we parse the version with a regex. We could set an `agent_major_version` variable once and use it throughout the recipe.

albertvaka: I thought about optimizing this, but ended up preferring the simplicity of this solution to the better performance of caching the result.

PR #666, opened 2019-12-12
Put monitors default configuration into a dir

kbogtob:
### What does this PR do?

When generating the default configuration for a monitor, using the monitor resource, the generated file is now put into a specific folder relevant to the monitored resource instead of just creating a file in `/etc/datadog-agent/conf.d`.

For example for mysql, instead of having one generated file at `/etc/datadog-agent/conf.d/mysql.yaml`, the generated file will be put into `/etc/datadog-agent/conf.d/mysql.d/conf.yaml`. 

Special thanks to @iancward who inspired this PR with his contribution on #558.

### Motivation

In reference to #558, the main idea is to allow users and client cookbooks to define several configuration files for their agent's integrations. Generating a file into a specific folder known by the agent will allow it to read the whole folder, thus allowing users to add more files into that specific folder.

### Known Problems

Today, the mesos integration is not working as the generated file in /etc/datadog-agent/conf.d/mesos.yaml is not interpreter as there is now two integrations, `mesos_slave` and `mesos_master` (and their relevant folders `mesos_slave.d` and `mesos_master.d`). This PR breaks the logic as it base itself on the fact that there is a folder having the integration's name. This breaks the mesos-integration kitchen tests.

I will fix the mesos integration very soon in another PR and fix the relevant kitchen tests.

### DIFF
--- a/resources/monitor.rb
+++ b/resources/monitor.rb
@@ -16,7 +16,7 @@ property :logs, [Array, nil], required: false, default: []
 action :add do
   Chef::Log.debug("Adding monitoring for #{new_resource.name}")
 
-  template ::File.join(config_file_path(new_resource.name), 'conf.yaml') do
+  template config_file_path(new_resource.name) do
     # On Windows Agent v5, set the permissions on conf files to Administrators.
     if node['platform_family'] == 'windows'
       if node['datadog']['agent6']
@@ -59,8 +59,17 @@ end
 
 def config_file_path(resource_name)
   if node['datadog']['agent6']
-    ::File.join(node['datadog']['agent6_config_dir'], 'conf.d', "#{resource_name}.d")
+    ::File.join(
+      node['datadog']['agent6_config_dir'],
+      'conf.d',
+      "#{resource_name}.d",
+      'conf.yaml'
+    )
   else
-    ::File.join(node['datadog']['config_dir'], 'conf.d', "#{resource_name}.d")
+    ::File.join(
+      node['datadog']['config_dir'],
+      'conf.d',
+      "#{resource_name}.yaml"
+    )
   end
 end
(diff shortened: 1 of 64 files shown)


albertvaka:
I wonder... If a user was running a previous version of the module and they upgrade, this means we will start creating the new file while the other one will stay there. I think we should delete the older managed file, to avoid problems.

resources/monitor.rb
```diff
@@ -53,10 +57,10 @@
...
 end
 
-def yaml_dir
+def config_file_path(resource_name)
   if node['datadog']['agent6']
-    ::File.join(node['datadog']['agent6_config_dir'], 'conf.d')
+    ::File.join(node['datadog']['agent6_config_dir'], 'conf.d', "#{resource_name}.d")
   else
-    ::File.join(node['datadog']['config_dir'], 'conf.d')
+    ::File.join(node['datadog']['config_dir'], 'conf.d', "#{resource_name}.d")
```
olivielpeau: Agent v5 doesn't support the per-check subdirectories, i.e. for integration `foo` Agent 5 will only read the config file in `<agent_config_dir>/conf.d/foo.yaml`.

So we'll need some logic specific to Agent 5, see what was done in #558 (and let's acknowledge and thank the initial contributor as well :) )

kbogtob:
> I wonder... If a user was running a previous version of the module and they upgrade, this means we will start creating the new file while the other one will stay there. I think we should delete the older managed file, to avoid problems.

That's done with 43cb999. What do you think?

resources/monitor.rb
kbogtob: My bad for the unique file for agent 5. It's done with 28cf699.

kbogtob:
I have to rebase the branch after the mesos integration PR.

PR #673, opened 2019-12-20
[chef][agent5] allow configuration of profiling option

truthbk:
knob for: https://github.com/DataDog/dd-agent/blob/6e33f4209b8bd8a3023d3795e8348bf37c580803/agent.py#L316

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -422,7 +422,7 @@ default['datadog']['autorestart'] = false
 
 # Run the agent in developer mode
 # Agent v5 only.
-default['datadog']['developer_mode']  = false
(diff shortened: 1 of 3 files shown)


attributes/default.rb
```diff
@@ -422,7 +422,11 @@
 
 # Run the agent in developer mode
 # Agent v5 only.
-default['datadog']['developer_mode'] = false
+default['datadog']['developer_mode']  = false
```
truthbk:
```suggestion
default['datadog']['developer_mode'] = false
```

truthbk:
Let's use `extra_config` instead! Closing!

PR #674, opened 2019-12-20
Add development guidelines

kbogtob:
### What does this PR do?

It adds some development guidelines in the README.md, it adds a CONTRIBUTING.md file and fixes the chef version in the temporary Gemfile so tests can be ran uniformly on any dev environment.

### Motivation

On boarding on the project might not seem easy if you are not used with Chef and Kitchen. Hence the addition of documentation to have more leads and hints on how you can work on the project.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,55 @@
+# Contributing to the Datadog Chef recipe
+
+First of all, thanks for contributing!
+
+This document provides some basic guidelines for contributing to this repository.
+To propose improvements, feel free to submit a PR.
+
+## Submitting issues
+
+  * If you think you've found an issue, please search the issue list to see if there's a matching existing issue.
+  * Then, if you find nothing, please open a Github issue.
+
+## Pull Requests
+
+Have you fixed a bug or written a new feature and want to share it? Many thanks!
+
+In order to ease/speed up our review, here are some items you can check/improve
+when submitting your PR:
+
+  * have a [proper commit history](#commits) (we advise you to rebase if needed).
+  * write tests for the code you wrote.
+  * preferably make sure that all unit tests pass locally and some relevant kitchen tests.
+  * summarize your PR with an explanatory title and a message describing your
+    changes, cross-referencing any related bugs/PRs.
+  * open your PR against the `master` branch.
+
+Your pull request must pass all CI tests before we will merge it. If you're seeing
+an error and don't think it's your fault, it may not be! [Join us on Slack][slack]
+or send  us an email, and together we'll get it sorted out.
+
+### Keep it small, focused
+
+Avoid changing too many things at once. For instance if you're fixing a recipe and at the same time adding some code refactor, it makes reviewing harder and the _time-to-release_ longer.
+
+### Commit Messages
+
+Please don't be this person: `git commit -m "Fixed stuff"`. Take a moment to
+write meaningful commit messages.
+
+The commit message should describe the reason for the change and give extra details
+that will allow someone later on to understand in 5 seconds the thing you've been
+working on for a day.
+
+If your commit is only shipping documentation changes or example files, and is a
+complete no-op for the test suite, please add **[skip ci]** in the commit message
+body to skip the build and give that slot to someone else who does need it.
+
+### Squash your commits
+
+Please rebase your changes on `master` and squash your commits whenever possible,
+it keeps history cleaner and it's easier to revert things. It also makes developers
+happier!
+
+
+[slack]: http://datadoghq.slack.com
--- a/Gemfile
+++ b/Gemfile
@@ -2,16 +2,14 @@ source 'https://rubygems.org'
 
 # Gems here are used to run `rake release`.
 
-group :release do
-  gem 'emeril'
-  gem 'rake'
-  gem 'chef', '~> 12.22'
-  gem 'foodcritic'
-  gem 'chefspec'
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -445,3 +445,78 @@ their `CHANGELOG.md` file in the [integrations-core repository](https://github.c
...
+To contribute, you will have to follow the contribution guide in [CONTRIBUTING.md](./CONTRIBUTING.md).
+
+### Dependencies
+
+First, you should use [bundler](https://bundler.io) and the provided Gemfile to install the development and release dependencies.
+
+After having installed bundler, you can run the following command to install gem in a vendored folder:
+
+```bash
+bundle install --path .bundle
```
albertvaka: CircleCI uses chefdk (that provides a virtual ruby environment like bundler), so this might mean that versions in the dev env are slightly different than in the CI env.

I think bundler is the go-to tool in the Ruby world, so I like advising to use `bundle` in the instructions, but maybe we should change the CI to also use it. What do you think? This is something we can do later.

I think the `--path .bundle` is the default and we can skip it.

Gemfile
```diff
@@ -5,7 +5,7 @@ source 'https://rubygems.org'
 group :release do
   gem 'emeril'
   gem 'rake'
-  gem 'chef', '~> 12.22'
+  gem 'chef', '= 14.10.9'
   gem 'foodcritic'
```
albertvaka: The CI runs with foodcritic 11.4.0, and installing latest actually gives a deprecation warning. I think we should pin that version here as well.

Gemfile
```diff
@@ -5,7 +5,7 @@ source 'https://rubygems.org'
 group :release do
   gem 'emeril'
   gem 'rake'
-  gem 'chef', '~> 12.22'
+  gem 'chef', '= 14.10.9'
```
albertvaka: The CI tests with chef 13.4, so maybe we should pin that version. Also, is there any reason to pin an exact minor and not use `~>`>

Also, note that emeryl is not compatible with chef > 12. But the right way to solve this is to get rid of emeryl and find a different way to do the releases... The current way is broken in so many ways anyway.

README.md
```diff
@@ -445,3 +445,78 @@ their `CHANGELOG.md` file in the [integrations-core repository](https://github.c
...
+bundle exec kitchen test <target>
+```
+
+So for example, if you want to test the CouchDB monitor on Ubuntu 16.04, you can run:
+
+```bash
+bundle exec kitchen test datadog-couchdb-ubuntu-1604-15
+```
+
+As there is a CouchDB recipe described in the `kitchen.yml`, as an Ubuntu platform with the 16.04 version.
```
albertvaka:
```suggestion
As there is a CouchDB recipe described in the `kitchen.yml`, and an Ubuntu platform with the 16.04 version.
```

README.md
```diff
@@ -445,3 +445,78 @@ their `CHANGELOG.md` file in the [integrations-core repository](https://github.c
...
+
+More information about kitchen on its [Getting Started](https://kitchen.ci/docs/getting-started/introduction/).
+
+### Development loop
```
albertvaka: We should mention that `kitchen login` is only for Linux. On Windows, you need to login graphically to the VM. We should also mention that the user/pass for the Windows VMs we use in kitchen is vagrant/vagrant.

CONTRIBUTING.md
```diff
@@ -0,0 +1,55 @@
+# Contributing to Datadog Agent
```
albertvaka:
```suggestion
# Contributing to the Datadog Chef recipe
```

Gemfile
```diff
@@ -5,7 +5,7 @@ source 'https://rubygems.org'
 group :release do
```
albertvaka: The :group should be removed.

README.md
kbogtob: Yeah we should definitely use bundler so the tests are run reproducibly. 

The option `--path .bundle` is not the default one sadly. The default one is installing in the user wide bundle gems folder. Generally in `$HOME/.bundle/gems`

Gemfile
kbogtob: With the version `12.22`, the rspec tests do not run and some tests are failing because of default values management (empty array in `12.22`, nil in most recent versions).
I've chosen the same version as the CI:
```bash
#!/bin/bash -eo pipefail
chef --version
Chef Development Kit Version: 3.8.14
chef-client version: 14.10.9
delivery version: master (9d07501a3b347cc687c902319d23dc32dd5fa621)
berks version: 7.0.7
kitchen version: 1.24.0
inspec version: 3.6.6
```

README.md
kbogtob: You're right! Thanks.

Gemfile
kbogtob: Good idea.

Gemfile
kbogtob: Done.

Gemfile
albertvaka: 👍 Checking again, it looks like we use 13 in Travis and 14 in Circle. When we unify the CIs we should make sure to test both versions (and probably 15 as well).

albertvaka: And yup, tests don't pass in 12. The reason 12 was pinned here was because of Emeryl.

kbogtob: Yup, I checked only on CircleCI. I get it now. Thank you!

PR #675, opened 2019-12-23
Automatically preppend "1:" and append "-1" to version num if needed

albertvaka:
On Debian-based and SUSE, it's no longer needed to pass the full version.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -32,15 +32,13 @@ default['datadog']['application_key'] = nil
 default['datadog']['agent_major_version'] = nil # nil to autodetect based on 'agent_version'
 
 # Agent Version
-# Default of `nil` will install latest version. On Windows, this will also upgrade to latest
-# This attribute accepts either a `string` or `hash` with the key as platform_name and value of package version
-# In the case of fedora and amazon linux, use platform_name of rhel
+# Default of `nil` will install latest version. On Windows, this will also upgrade to latest.
+# This attribute takes a `string` with a version number. For compatibiliy reasons it can also take a
+# hash with the platform_name as key and the version as value.
+# Starting with version 4.1.0, you no longer need to pass the "1:" prefix and "-1" suffix in versions
+# for Debian-based and SUSE.
 # Example:
-# default['datadog']['agent_version'] = {
-#  'rhel' => '5.9.0-1',
-#  'windows' => '5.9.0',
-#  'debian' => '1:5.9.0-1'
-# }
+# default['datadog']['agent_version'] = '7.16.0'
 default['datadog']['agent_version'] = nil # nil to install latest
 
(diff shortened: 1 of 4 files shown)


attributes/default.rb
```diff
@@ -32,15 +32,13 @@
...
-# This attribute accepts either a `string` or `hash` with the key as platform_name and value of package version
-# In the case of fedora and amazon linux, use platform_name of rhel
+# Default of `nil` will install latest version. On Windows, this will also upgrade to latest.
+# This attribute takes a `string` with a version number. For compatibiliy reasons it can also take a
+# hash with the platform_name as key and the version as value.
+# Starting with version 4.1.0, you no longer need to pass the "1:" prefix and "-1" suffix in versions
+# for Debian-based and SUSE.
 # Example:
-# default['datadog']['agent_version'] = {
-#  'rhel' => '5.9.0-1',
```
KSerrania: What about RHEL? We used to add the `-1` revision in the version name, is that not needed anymore? Or do we need to add another case for it?

albertvaka: I'll test.

libraries/recipe_helpers.rb
```diff
@@ -12,6 +12,11 @@ def agent_version(node)
           end
           dd_agent_version = dd_agent_version[platform_family]
         end
+        if node['platform_family'] == 'suse' || node['platform_family'] == 'debian' then
```
KSerrania:
```suggestion
        if node['platform_family'] == 'suse' || node['platform_family'] == 'debian'
```

libraries/recipe_helpers.rb
```diff
@@ -12,6 +12,11 @@ def agent_version(node)
...
+          if !dd_agent_version.nil? && dd_agent_version.match(/^[0-9]+\.[0-9]+\.[0-9]+((?:~|-)[^0-9\s-]+[^-\s]*)?$/)
+            dd_agent_version = "1:" + dd_agent_version + "-1"
```
KSerrania:
```suggestion
            dd_agent_version = '1:' + dd_agent_version + '-1'
```

attributes/default.rb
albertvaka: On CentOS 6 it seems to work without the added `-1`.

KSerrania: Nice. I wonder why we added that `-1` in all tools where we install RHEL though 🤔

albertvaka: Works on Fedora 21 as well, I think the `-1` in the example was just a typo.

albertvaka: In what other tool were we adding `-1`?

albertvaka: It seems to work without the `-1` everywhere I tried.

PR #676, opened 2019-12-26
Use bundler on CircleCI and fix dependency versions

kbogtob:
### What does this PR do?

Update the Gemfile to fix versions and a Gemfile.lock so that versions could be fixed and stable. Use the Gemfile on CircleCI with bundler to make pipelines more stable.

### Motivation

Make the tests ran on developer workstations similar to the ones on the CI so that the project could win in consistency and reliability.


Gemfile
```diff
@@ -1,15 +1,18 @@
...
+group :release do
+  gem 'emeril'
+end
 
-gem 'emeril'
 gem 'rake'
 gem 'chef', '= 14.10.9'
-gem 'foodcritic', '= 11.4.0'
+gem 'foodcritic', '~> 11.4.0'
+gem 'cucumber-core', '~> 3.2.1'
```
kbogtob: As foodcritic is laxist on its dependencies, if we don't fix the cucumber version, it breaks the foodcritic requirements.

kbogtob:
PR canceled by #677

PR #678, opened 2019-12-30
[monitor] fix issues with directory creation on windows 

truthbk:
There was a bit of copy/paste going on here, so I'm sure this can be fixed in a cleaner way. I'll let you guys take the PR over, but it should be fixed as this is currently broken on windows.

### DIFF
--- a/resources/monitor.rb
+++ b/resources/monitor.rb
@@ -19,13 +19,7 @@ action :add do
   if Chef::Datadog.agent_major_version(node) != 5
     directory ::File.join(yaml_dir, "#{new_resource.name}.d") do
       if node['platform_family'] == 'windows'
-        if Chef::Datadog.agent_major_version(node) > 5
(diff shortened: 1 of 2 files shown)


resources/monitor.rb
```diff
@@ -18,9 +18,19 @@
 
   if Chef::Datadog.agent_major_version(node) != 5
     directory ::File.join(yaml_dir, "#{new_resource.name}.d") do
-      owner 'dd-agent'
-      group 'dd-agent'
-      mode '755'
+      if node['platform_family'] == 'windows'
+        if Chef::Datadog.agent_major_version(node) > 5
```
KSerrania: I think this check is not needed, we're already in the `Chef::Datadog.agent_major_version(node) != 5` branch

truthbk: Yes, this is exactly what I was talking about. I hadn't even bother checking the logic, thanks.

PR #679, opened 2019-12-30
Add standalone release process

kbogtob:
### What does this PR do?

It adds a Rake task to release the cookbook on the Chef supermarket. And it gets rid of emeril.

### Motivation

Getting rid of emeril so we can update our dependencies. Add an up to date release process based on knife supermarket share instead of the cookbook site share (that's what uses Emeril) that is deprecated and that will soon be closed for good.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,9 +1,5 @@
 source 'https://rubygems.org'
 
-group :release do
-  gem 'emeril'
-end
-
 gem 'rake'
 gem 'chef', '= 14.10.9'
 gem 'foodcritic', '~> 11.4.0'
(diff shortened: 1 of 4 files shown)


kbogtob:
I released this cookbook with it: https://supermarket.chef.io/cookbooks/ddrelease

kbogtob:
While looking at the files bundled, there's temporary development files. I thought the chefignore might be sufficient but it may not be a good idea to release directly the working dir.
I'll add a git clone into temporary directory process before applying the release process.

albertvaka:
Something similar happens with Puppet and we solved it by documenting the extra step of making a fresh clone. +1 to automating it, though.

PR #681, opened 2020-01-08
Make kitchen docker tests compatible with agent >5

kbogtob:
### What does this PR do?
Added kitchen docker tests using the agent 6 and 7 on every OSes. Agent > 5 does not use supervisord but system specific init runners like system 5 or systemd. For systemd OSes, it use specific homemade images simulated systemd. For system5 OSes, the PR https://github.com/DataDog/datadog-agent/pull/4691 fix the system5 init.d scripts for compability with docker.

### Motivation

We will be able to run chef kitchen docker tests on the CI for agent 5 but also for agent 6 and 7.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -96,19 +96,11 @@ task :circle do
   end
 
   def command
-    commands = []
-
-    kitchen_config.platforms.sort_by(&:name).each_with_index do |platform, index|
+    kitchen_config.instances.sort_by(&:name).each_with_object([]).with_index do |(instance, commands), index|
       next unless index % total_workers == current_worker
-      # Escape the platform name to somehting the CLI will understand.
-      # TODO: This could likely be pulled from kitchen_config.instances somehow
-      name = platform.name.delete('.')
 
-      # Scope the suites to only execute against the Agent+handler installer suites.
-      commands.push "kitchen verify dd-agent-handler-#{name}"
-    end
-
-    commands.join(' && ')
+      commands << "kitchen verify #{instance}"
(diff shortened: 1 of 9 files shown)


Rakefile
```diff
@@ -96,19 +96,11 @@ task :circle do
...
-      # Escape the platform name to somehting the CLI will understand.
-      # TODO: This could likely be pulled from kitchen_config.instances somehow
-      name = platform.name.delete('.')
-
-      # Scope the suites to only execute against the Agent+handler installer suites.
-      commands.push "kitchen verify dd-agent-handler-#{name}"
-    end
 
-    commands.join(' && ')
+      commands << "kitchen verify #{instance}"
```
albertvaka: 🍰

kitchen.docker.yml
```diff
@@ -15,21 +15,38 @@ platforms:
...
     - recipe[datadog::dd-agent]
+    - recipe[datadog::dd-handler]
   attributes:
     datadog:
-      # TODO(remy): run those tests with Agent v6 recipe.
-      agent_major_version: 5
+      agent_major_version: <%= agent_major_version %>
       api_key: somenonnullapikeythats32charlong
       application_key: alsonotnil
+      chef_handler_enable: false
```
albertvaka: Is this `chef_handler_enable ` needed?

kbogtob: It's because last version of chef-handler had an issue due to the dogapi. I'll re-enable it soon.

PR #687, opened 2020-01-29
Add tags to directory.yaml.erb

dimier:
Tags key was missed in directory.yaml.erb


kbogtob:
Thanks for your contribution @dimier ! The PR looks good to me, wdyt @albertvaka ?

PR #688, opened 2020-02-05
Use new S3 URL format

KSerrania:
### What does this PR do?

Switch S3 URLs to virtual-hosted style.

### Motivation

Comply with the new standard. Even if old buckets should not be affected, let's be consistent in the path style we use.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -158,7 +158,7 @@ default['datadog']['yumrepo_gpgkey'] = "#{yum_protocol}://yum.datadoghq.com/DATA
 default['datadog']['yumrepo_proxy'] = nil
 default['datadog']['yumrepo_proxy_username'] = nil
 default['datadog']['yumrepo_proxy_password'] = nil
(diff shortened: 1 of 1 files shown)


Xorima:
This should be clearly flagged for people using proxys btw so they do not get caught out ...

albertvaka:
> This should be clearly flagged for people using proxys btw so they do not get caught out ...

This might be a good enough reason not to do this, actually :/

Xorima:
> > This should be clearly flagged for people using proxys btw so they do not get caught out ...
> 
> This might be a good enough reason not to do this, actually :/

While I agree with you, I would prefer to have a more specific url to whitelist than not, and `ddagent-windows-stable.s3.amazonaws.com` is a lot better than being forced to whitelist the whole of s3 just because datadog uses it, I guess it's your choice? 

You could always log a warning about it will be changed to `ddagent-windows-stable.s3.amazonaws.com` in the next major cookbook release so users have time to react

KSerrania:
I'll be closing this for now. We'll open a new PR later when we decide to switch to the new `windows-agent.datadoghq.com` URL, which will be clearer than using an S3 URL.

PR #690, opened 2020-02-24
Force reinstallation of agent when downgrading to old version on Windows

kbogtob:
### What does this PR do?

It reinstalls the agent, only if the targeted agent_version is lower than the currently installed version. Also, it skips the reinstallation on chef versions lower than 14 as it does not correctly uninstall the agent.

### Motivation

The agent installer does not support downgrading since the fix for the windows incident on 6.14. By allowing to explicitly uninstall and then reinstall the agent when a lower version of the agent is targeted, we are allowing our users to downgrade.

### DIFF
--- a/libraries/recipe_helpers.rb
+++ b/libraries/recipe_helpers.rb
@@ -104,6 +104,10 @@ class Chef
         end
 
         def chef_version_can_uninstall?
+          # Chef versions previous to 14 cannot correctly uninstall the agent
+          # because they cannot correctly fetch the registry keys of 64 bits
+          # applications for uninstallation so we are only using the downgrade
+          # feature on chef >= to 14
           Gem::Requirement.new('>= 14').satisfied_by?(Gem::Version.new(Chef::VERSION))
         end
       end
--- a/libraries/recipe_helpers.rb
+++ b/libraries/recipe_helpers.rb
@@ -69,5 +69,44 @@ class Chef
         end
       end
     end
+
+    module WindowsInstallHelpers
+      WIN_BIN_PATH = 'C:/Program Files/Datadog/Datadog Agent/bin/agent'.freeze
+
+      class << self
+        def must_reinstall?(node)
+          current_version = fetch_current_version
+          target_version = requested_agent_version(node)
+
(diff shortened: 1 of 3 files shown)


recipes/_install-windows.rb
```diff
@@ -19,6 +19,33 @@
...
+  end
+end
+
+def must_reinstall?
+  current_version = fetch_current_version
+
+  return false unless current_version && requested_agent_version
+  return false unless requested_agent_version < current_version
+
+  true
```
albertvaka:
```suggestion
  requested_agent_version < current_version
```
And remove the second `return false unless ...`.

recipes/_install-windows.rb
```diff
@@ -19,6 +19,36 @@
...
+
+  Gem::Version.new(match_data[1]) if match_data
+end
+
+def requested_agent_version
+  version = Chef::Datadog.agent_version(node)
+  Gem::Version.new(version) if version
+end
+
+def chef_version_uninstall_correctly?
```
albertvaka: Can you add a comment explaining why we need to check this? That is: Chef < 14 not checking the right registry entry.

libraries/recipe_helpers.rb
```diff
@@ -69,5 +69,47 @@ def run_state_or_attribute(node, attribute)
...
+          version = Chef::Datadog.agent_version(node)
+          return nil unless version
+
+          cleaned = version.scan(/\d+\.\d+\.\d+/).first
+          Gem::Version.new(cleaned) if cleaned
+        end
+
+        def chef_version_can_uninstall?
+          # Chef versions previous to 14 cannot correctly uninstall the agent
+          # because they cannot correctly fetch the registry keys for uninstallation
```
albertvaka:
```suggestion
          # because they cannot correctly fetch the registry keys of 64bit applications for uninstallation
```

kbogtob: Totally!

kbogtob:
There's something I just thought about, the cookbook supports fixing the agent version with a git hash. What should we do in that case? Right now, it won't be considered by the patch and no downgrade will occur.

PR #691, opened 2020-02-24
Set windows installer as sensitive resource

julien-lebot:
The current installation method can leak credentials in the Chef events, because it is not marked as sensitive.


kbogtob:
It looks good to me but I'll check that it correctly removes the logs when specifying a username and a password.

recipes/_install-windows.rb
```diff
@@ -104,6 +104,16 @@
   notifies :remove, 'package[Datadog Agent removal]', :immediately
 end
 
+windows_env 'DDAGENTUSER_NAME' do
+  value "".concat(Chef::Datadog.ddagentuser_name(node)) if Chef::Datadog.ddagentuser_name(node)
+  sensitive true
+end
+
+windows_env 'DDAGENTUSER_PASSWORD' do
+  value "".concat(Chef::Datadog.ddagentuser_password(node)) if Chef::Datadog.ddagentuser_password(node)
```
kbogtob: Why using `"".concat(Chef::Datadog.ddagentuser_name(node))`?

julien-lebot: Yeah I can use the value directly, gotcha

recipes/_install-windows.rb
```diff
@@ -127,3 +137,11 @@
...
   end
 end
+
+windows_env 'DDAGENTUSER_NAME' do
+  action :delete
+end
+
+windows_env 'DDAGENTUSER_PASSWORD' do
+  action :delete
+end
```
julien-lebot: Note: In case the install fails, those steps are never taken. Hence the username/password used will stay stored in the environment variables. We'll need to make sure this is not an issue for customers before releasing.

image

Issue #692, opened 2020-02-24
Incorrect comments re. value of skip_ssl_validation


jaxi:
https://github.com/DataDog/chef-datadog/blob/e0451e78b747e307e609f5bb1f4ed9d45b0e51fa/attributes/default.rb#L250 suggests that `skip_ssl_validation` "accepted values 'yes' or 'no'", however in [DD agent proxy config documentation](https://docs.datadoghq.com/agent/proxy/?tab=agentv6v7#datadog-agent-configuration), the value is expected to be a boolean

```
skip_ssl_validation: true
```

I initially gave `default['datadog']['web_proxy']['skip_ssl_validation'] = 'yes'` a go, which causes TLS handshake error with the proxy server, on the other hand `default['datadog']['web_proxy']['skip_ssl_validation'] = true` works as expected.

Notes that correct this also potentially leads https://github.com/DataDog/chef-datadog/blob/e0451e78b747e307e609f5bb1f4ed9d45b0e51fa/templates/default/datadog.conf.erb#L14 to be change into

```
<% unless node['datadog']['web_proxy']['skip_ssl_validation'].nil? -%>
...
```

jaxi:
PS: if it is regarded as a real issue I'm more than happy to submit a PR to fix it up :)

kbogtob:
Thanks for raising the issue! I can confirm the issue you are describing for recent agents, but I suppose the yes/no values were used by older agent versions. I'll check and confirm you if want to work on a PR :).

kbogtob:
The values 'yes' / 'no' were indeed needed by agent v5 (https://docs.datadoghq.com/agent/proxy/?tab=agentv5#datadog-agent-configuration). So the cookbook must support both.
@jaxi could you propose a PR for that? Your contribution would be greatly appreciated.

albertvaka:
Agent 5 can also take true/false [1], so we should simply change it to print the boolean.

[1] https://github.com/DataDog/dd-agent/blob/master/config.py#L598

jaxi:
I see. In this case a sane approach might simply be change the default attribute into

```
default['datadog']['web_proxy']['skip_ssl_validation'] = false
```

Right now GH keep throwing me 500 on forking the repo, will submit a PR once it comes back to normal :)

jaxi:
@albertvaka @kbogtob pull request ready in https://github.com/DataDog/chef-datadog/pull/693. Please let me know what you think.

albertvaka:
[...]

PR #693, opened 2020-02-25
Default skip_ssl_validation into false

jaxi:
Related issue: https://github.com/DataDog/chef-datadog/issues/692

Modified comments on skip_ssl_validation that suggests true/false is supported value to be populated into datadog config, which is compatible across agent v5, v6, and v7

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -247,7 +247,7 @@ default['datadog']['web_proxy']['host'] = nil
 default['datadog']['web_proxy']['port'] = nil
 default['datadog']['web_proxy']['user'] = nil
 default['datadog']['web_proxy']['password'] = nil
-default['datadog']['web_proxy']['skip_ssl_validation'] = nil # accepted values 'yes' or 'no'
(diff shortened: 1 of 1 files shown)


attributes/default.rb
```diff
@@ -247,7 +247,7 @@
 default['datadog']['web_proxy']['port'] = nil
 default['datadog']['web_proxy']['user'] = nil
 default['datadog']['web_proxy']['password'] = nil
-default['datadog']['web_proxy']['skip_ssl_validation'] = nil # accepted values 'yes' or 'no'
+default['datadog']['web_proxy']['skip_ssl_validation'] = false
```
albertvaka: Can we keep it nil here and just change the comment? This way it won't appear in the configuration if not set, thus using the default.

jaxi: Thanks for the feedback. I've addressed the comment.

albertvaka:
Thanks a lot for finding this! Merging :)

PR #694, opened 2020-02-27
Add cleanup for Windows env vars

julien-lebot:
When the `windows_package[Datadog Agent]` fails, it does not cleanup the environment variables containing the credentials for the installer.
This PR aims to fix that by setting a hanlder that runs after the recipe.


recipes/_install-windows.rb
```diff
@@ -141,11 +154,3 @@
...
   end
 end
-
-windows_env 'DDAGENTUSER_NAME' do
-  action :delete
-end
-
-windows_env 'DDAGENTUSER_PASSWORD' do
-  action :delete
-end
```
albertvaka: I think we can still keep these deletes, so the env vars are cleaned earlier than at `run_completed` or `run_failed`. 

Eg: if a run takes long and you kill it, `on_failed` might not get called, so it's better to delete the env vars earlier when we know we don't need them anymore.

libraries/windows_cleanup.rb
```diff
@@ -0,0 +1,15 @@
...
+module Windows
+  class Helper
+    def do_cleanup(context)
+      Chef::Log.info 'Cleanup started.'
+      resource = context.resource_collection.lookup("windows_env[DDAGENTUSER_NAME]")
+      resource.run_action(:delete) if resource
+      resource = context.resource_collection.lookup("windows_env[DDAGENTUSER_PASSWORD]")
+      resource.run_action(:delete) if resource
+      Chef::Log.info 'Cleanup finished.'
+    end
```
albertvaka: I would go for compact code and keep this helper function in `recipes/_install-windows.rb` instead of having several small files, but no strong opinion in case you prefer it this way.

PR #703, opened 2020-03-10
Test specs against more rubies

kbogtob:
### What does this PR do?

Add some circle ci jobs to run specs on different version of rubies and fix dependencies that are compatible with every versions.

### Motivation

Avoid issues with older rubies.


.circleci/config.yml
```diff
@@ -1,16 +1,16 @@
 version: 2
 jobs:
-  test:
+  complete-specs:
```
albertvaka: This name is not very descriptive: All (complete) spec tests are run by the version-specific jobs as well, but this one runs other stuff in addition to spec (style check, etc... not sure exactly what else).

kbogtob: What about `full-suite`?

albertvaka: 👍

PR #704, opened 2020-03-16
Allow configuration of logs agent

Nibons:
[FEATURE] Allow configuration of logs_agent via `datadog['logs_agent_config']`


attributes/default.rb
```diff
@@ -354,6 +354,11 @@
 # * `false` to explicitly disable it
 # Leave it to `nil` to let the agent's default behavior decide whether to run the log-agent
 default['datadog']['enable_logs_agent'] = nil
+default['datadog']['logs_agent_config'] = nil
+# {
+#   'use_http' => false,
+#   'use_compression' => true
+# }
```
albertvaka:
```suggestion
# Here you can specify any settings that should end up under 'logs_config' in datadog.yaml, eg:
# default['datadog']['logs_agent_config'] = {
#   'use_http' => false,
#   'use_compression' => true
# }
default['datadog']['logs_agent_config'] = nil
```
Moving example above the setting for consistency.

albertvaka:
👍Thanks! Merging 😃

PR #707, opened 2020-04-22
Add installation info metadata support

kbogtob:
### What does this PR do?

It adds the installation info metadata to the chef cookbook so the agent returns that it was installed by the triple chef/chef version/cookbook version

### Motivation.

Allow customers to track their chef installation. Allows us to make stats on installation methods.

### DIFF
--- a/recipes/install_info.rb
+++ b/recipes/install_info.rb
@@ -51,7 +51,7 @@ template flare_path do
 
   variables(
     install_method: "chef-#{Chef::VERSION}",
-    install_method_version: "datadog-cookbook-#{Chef::Datadog.cookbook_version(run_context)}"
+    install_method_version: "datadog_cookbook-#{Chef::Datadog.cookbook_version(run_context)}"
   )
   sensitive true
 end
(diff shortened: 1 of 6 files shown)


recipes/install_info.rb
```diff
@@ -0,0 +1,57 @@
...
+  else
+    owner 'dd-agent'
+    mode '600'
+  end
+
+  source 'install_info.yaml.erb'
+
+  variables(
+    install_method: "chef-#{Chef::VERSION}",
+    install_method_version: "datadog-cookbook-#{Chef::Datadog.cookbook_version(run_context)}"
```
albertvaka, line 54: Maybe we want to call it `datadog_cookbook` so the dash `-` separates name from version.

templates/default/install_info.yaml.erb
```diff
@@ -0,0 +1,6 @@
+<%= JSON.parse(({
```
albertvaka, line 1: `JSON.parse().to_json.to_yaml` is this the way to print a yaml? 🤔

recipes/install_info.rb
kbogtob: Woops! Good catch

templates/default/install_info.yaml.erb
kbogtob: C/p from other template, I'll improve

PR #708, opened 2020-04-24
[cassandra] Update the list of metrics to collect from jmx

k2v:
This PR updates the list of metrics to collect from JMX based on https://github.com/DataDog/integrations-core/blob/master/cassandra/datadog_checks/cassandra/data/metrics.yaml


albertvaka:
Why are we duplicating this file? Can't we use the one from the Agent in Chef?

PR #709, opened 2020-04-24
Fix datadog_monitor action :delete not using A6's path

albertvaka:
- Use the right path for `:delete` depending on the Agent version
- Add basic tests both for `:add` and `:delete` actions.

Unfortunately to test custom resources like `datadog_monitor` you need to create a test cookbook. Fortunately we already had one, although for some reason it was gitignored (but commited). This also unignores it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,6 @@ Gemfile*.lock
 
 # Berkshelf
 Berksfile.lock
-cookbooks/
 /.ruby-version
 
 # Test-Kitchen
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -28,7 +28,7 @@ Gemfile*.lock
 
 # Berkshelf
 Berksfile.lock
-cookbooks/
+/cookbooks/
```
kbogtob, line 31: why do we still want to keep `<root>/cookbooks`?

albertvaka: I assumed this was the original intent of the line, but I'm not sure if we need it. Just keeping changes to the minimum.

albertvaka: I'll delete it and if we see we need it we can add it again.

PR #711, opened 2020-04-30
Do not crash if agent_version doesn't match regex

albertvaka:
If there was no match, we would call `.catpures` on nil. Warn and continue.


kbogtob:
I like it, it's a safer solution.LGTM

PR #716, opened 2020-05-15
fix link to example conf yaml

gml3ff:
Link to example conf yaml in the win32 event log recipe is broken. 

This should fix it.


albertvaka:
Already done here: https://github.com/DataDog/chef-datadog/issues/715

PR #717, opened 2020-05-26
Add support of IOT agent flavor

kbogtob:
### What does this PR do?

It adds the base logic to support agent flavors. And it adds the support of the IOT agent.

### Motivation

Allowing users to install specific flavors of the agent.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -40,8 +40,8 @@ default['datadog']['agent_major_version'] = nil # nil to autodetect based on 'ag
 # Example:
 # default['datadog']['agent_version'] = '7.16.0'
 default['datadog']['agent_version'] = nil # nil to install latest
-# Agent flavor, acceptable values are "agent", iot-agent"
-default['datadog']['agent_flavor'] = 'agent' # "agent" to install default agent
+# Agent flavor to install, acceptable values are "agent", "iot-agent"
+default['datadog']['agent_flavor'] = 'agent' # "agent" to install the datadog-agent package
 
 # Allow override with `upgrade` to get latest (Linux only)
 default['datadog']['agent_package_action'] = 'install'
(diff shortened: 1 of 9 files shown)


attributes/default.rb
```diff
@@ -40,6 +40,8 @@
 # Example:
 # default['datadog']['agent_version'] = '7.16.0'
 default['datadog']['agent_version'] = nil # nil to install latest
+# Agent flavor, acceptable values are "agent", iot-agent"
```
mx-psi, line 43: nit: on the Agent [we use](https://github.com/DataDog/datadog-agent/blob/47c81996bc67aaf38b14f3cca34376dca6877fac/pkg/config/config.go#L61-L66) `agent` and `iot_agent` for the flavors, maybe we could use those values here too for consistency? I am not sure if the other ones are user-facing though, but just in case they ever are.

albertvaka: I like `iot-agent` better because it matches the name of the package. The `flavor` names in the Agent code are to identify them on our side, but shouldn't be user-facing.

attributes/default.rb
```diff
@@ -40,6 +40,8 @@
...
+default['datadog']['agent_flavor'] = 'agent' # "agent" to install default agent
```
albertvaka, line 44:
```suggestion
default['datadog']['agent_flavor'] = 'agent' # "agent" to install the datadog-agent package
```

attributes/default.rb
albertvaka, line 43:
```suggestion
# Agent flavor to install, acceptable values are "agent", "iot-agent"
```

KSerrania:
Would it be easier to set `agent_flavor`  to the package name (ie. `datadog-agent` / `datadog-iot-agent` instead of `agent` / `iot-agent`)? It would also make it behave the same way the install script does: https://github.com/DataDog/datadog-agent/blob/669abd39dae58460e3cc680c3af5a83bb763311d/cmd/agent/install_script.sh#L102-L105

albertvaka:
+1 to Kylian, it will also ease adding support for flavors that don't start with "datadog-" like dogstatsd.

attributes/default.rb
kbogtob: Great, thanks Albert

PR #718, opened 2020-05-29
Do not crash if the agent is not running

albertvaka:
Fixes `NoMethodError: undefined method '[]' for nil:NilClass` if `agent status` doesn't return anything useful (eg: if the agent is not running).


albertvaka:
The whole `recipe_helpers.rb` file is untested actually :(

I don't know how to test it in isolation, though, since the rspec tests we have test the results of a a full chef run, not of individual functions. Any idea?

PR #724, opened 2020-06-24
Refactor kitchen yaml

julien-lebot:
Refactor the kitchen.yaml so that the platforms are defined in common anchors at the beginning of the file.
Many tests could not run because omnitruck.chef.io/ was returning a 404 - so I implemented some logic to check that the platform is supported before creating a platform entry. Another option was to use a static list, but that would require a lot of manual checks to make sure the platforms are still supported.
Also, fix multiple anchor definitions of `DATADOG`.

### DIFF
--- a/kitchen.yml
+++ b/kitchen.yml
@@ -48,7 +48,6 @@ require "uri"
     - <%= platform_version %>-<%= chef_version %>
   <% platforms_combo[platform_version] = chef_version
   end
-  puts platforms_combo
 end %>
 
 .DATADOG: &DATADOG
--- a/kitchen.yml
+++ b/kitchen.yml
@@ -82,6 +82,9 @@ windows_chef_versions.each do |chef_version|
 %>
 
 - name: <%= windows_platform %>-<%= chef_version %>
+  driver:
+    gui: false
+    linked_clone: true
   driver_config:
     box: mwrock/Windows2012R2  # lightweight windows box
     require_chef_omnibus: <%= chef_version %>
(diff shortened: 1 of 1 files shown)


kitchen.yml
```diff
@@ -16,17 +17,54 @@ platforms:
...
+<% end %>
+
+.LINUX_PLATFORMS: &LINUX_PLATFORMS
+<% test_platforms.product(chef_versions).each do |platform_version, chef_version| 
+  platform_name = platform_version[/(.*)-(.*)/, 1]
+  platform_version = platform_version[/(.*)-(.*)/, 2]
+  uri = URI.parse("https://omnitruck.chef.io/stable/chef/metadata?v=#{chef_version}&p=#{platform_name}&pv=#{platform_version}&m=x86_64")
+  puts "Checking #{uri}"
+  resp = Net::HTTP.get_response(uri)
+  puts resp.code == "200"
```
albertvaka, line 46: This is quite spammy (two lines per os*chef) and, the first time I ran it, quite slow (subsequent runs were faster for some reason but still not instant) which is quite annoying when you are just running a `kitchen list` or another command that should be fast.

Since the output will be always the same, can we hardcode the list of supported os*chef products instead of calculating it every time? Or calculate it once and cache it, if you prefer (but it's more work, not sure it's worth it).

julien-lebot: The reason I did this is because the initial static list was not correct and it was time consuming to run the tests (and spawn the machines) just to find out the platform was not specified. Thus I thought it was worth the time spent at the beginning of the tests to actually check that the test was going to be able to run before running it.
I think it should be possible to create a cache file and use it instead.

PR #730, opened 2020-07-13
Adds generated recipes for missing checks

albertvaka:
Fixes `bundle exec rake create_integration_monitor` introduced in #720
and adds 54 recipes created with it.

I didn't include checks that already have recipes since in some cases it
causes behavior changes or renames, so we might want to wait for a major
release bump to do that. Except for riak, since it looked safe to replace.

Left out `zk`, `elastic` and `couch` even though they are "new" recipes
because they aren't actually new: the current ones are named `zookeeper`,
`elasticsearch` and `couchdb`, which doesn't match the generated filenames.


KSerrania (approved):
The logic change looks good to me. I only have one question:

> I didn't include checks that already have recipes since in some cases it
causes behavior changes or renames, so we might want to wait for a major
release bump to do that. Except for riak, since it looked safe to replace.

This PR adds:
```
  init_config node['datadog']['riak']['init_config']
...
  use_integration_template true
```

Is this change safe because it doesn't actually change the behavior of the integration?

albertvaka:
Yes, the previous recipe used the riak template [1] which essentially results in the same as the default template [2] except that the riak template only took into account the `url` field within the instances array. Now it will still support that field, but you can add all the supported fields as well.

[1] https://github.com/DataDog/chef-datadog/blob/master/templates/default/riak.yaml.erb
[2] https://github.com/DataDog/chef-datadog/blob/master/templates/default/integration.yaml.erb

PR #732, opened 2020-08-07
Warn when the Agent will start despite being disabled

### LINKED ISSUE #447: Possible to install agent without starting it up?
We build AMI images for all of our apps. I want to speed up the Chef run on firstboot once the ec2 instance comes up. I notice installing the datadog agent takes a bit of time so I wanted to move that piece into the Packer build. How can I go about installing the agent but not have it start and send metrics to Datadog?

albertvaka:
Since the trace and process agents depend on the main agent, disabling the later without disabling the former won't have the expected behavior.

Fixes #447

### DIFF
--- a/recipes/dd-agent.rb
+++ b/recipes/dd-agent.rb
@@ -44,6 +44,14 @@ else
   include_recipe 'datadog::_install-linux'
 end
 
+if !node['datadog']['agent_enable'] && node['datadog']['enable_process_agent']
+  Chef::Log.warn("'agent_enable' is set to 'false', but 'enable_process_agent' is set to 'true'. This will cause the datadog-agent to start, since the process-agent depends on it.")
+end
+
+if !node['datadog']['agent_enable'] && node['datadog']['enable_trace_agent']
+  Chef::Log.warn("'agent_enable' is set to 'false', but 'enable_trace_agent' is set to 'true'. This will cause the datadog-agent to start, since the trace-agent depends on it.")
+end
+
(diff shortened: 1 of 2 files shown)


recipes/dd-agent.rb
```diff
@@ -44,6 +44,16 @@
   include_recipe 'datadog::_install-linux'
 end
 
+
+if node['datadog']['agent_enable'] && !node['datadog']['enable_process_agent']
```
mx-psi, line 48: Shouldn't it be the other way around? We want to log if the Agent is **not** enabled and the Process Agent **is** enabled. Maybe I am reading it incorrectly 🤔 
```suggestion
if !node['datadog']['agent_enable'] && node['datadog']['enable_process_agent']
```

recipes/dd-agent.rb
```diff
@@ -44,6 +44,16 @@
...
+  Chef::Log.warn("'agent_enable' is set to 'false', but 'enable_process_agent' is set to 'true'. This will cause the datadog-agent to start, since the process-agent depends on it.")
+end
+
+if node['datadog']['agent_enable'] && !node['datadog']['enable_trace_agent']
```
mx-psi, line 52: ditto
```suggestion
if !node['datadog']['agent_enable'] && node['datadog']['enable_trace_agent']
```

recipes/dd-agent.rb
albertvaka: My mistake, you are reading it right.

PR #734, opened 2020-08-12
Add third-party integrations support

julien-lebot:
# What does this PR do?
Adds a third_party option to the `integration` resource to specify that the check to install is a third-party check (defined in https://github.com/DataDog/integrations-extras).

# Motivation
Installing third-party integrations is supported since Agent 6.21.0/7.21.0.

# Additional notes
Adds chefspec test for the `integration` resource.


libraries/test_helpers.rb
```diff
@@ -0,0 +1,12 @@
+class Mock
+  class ShellCommandResult
+    @stdout = ''
+    def initialize(stdout)
+      @stdout = stdout
+    end
+
+    def stdout
+      @stdout
+    end
```
mx-psi, line 10: CI linter suggests doing
```suggestion
     attr_accessor :stdout
```
I think

Issue #683, opened 2020-01-21
dd-agent recipe is not idempotent


limitusus:
dd-agent recipe always evaluates `ruby_block 'include system-probe'` even after the first run, with rendered template placed on the host.
Chef recipes all should be idempotent, so most of conditionals should be evaluated as guard clauses.
What do you think of moving conditionals in `ruby_block`'s `only_if `clause or `template`'s `only_if` ?

I'm afraid of you need to fix current sysprobe tests somehow, because the current tests assumes sysprobe implies dd-agent was already installed.

julien-lebot:
Hey @limitusus, thanks for reporting this issue. We have released [v4.5.0](https://github.com/DataDog/chef-datadog/releases/tag/v4.5.0) of the cookbook which contains a fix that should address it (see #736).
Could you please let us know if the changes are satisfactory ? Thanks 👍

limitusus:
Sorry for late reply. I confirmed it fixed. Thank you so much 😄

PR #736, opened 2020-09-01
Make dd-agent recipe idempotent

julien-lebot:
### Summary
This PR aims to address issue #683, the `dd-agent` recipe was always running the `ruby_block` which was causing a resource update at each run (With message for example 'Chef Infra Client finished, 1/26 resources updated in 02 seconds').

Given that the `system-probe` is a dependency of the main agent, it's safe to always include the `system-probe` recipe from the `dd-agent` recipe, unless it's explicitly not supported or not enabled.

An other issue mentioned in #683 concerns the creation of the `system-probe.yaml` template on the host regardless if `system-probe` is enabled or not. However the attribute `['datadog']['system_probe']['manage_config']` is meant to govern this. Nevertheless, I added a clause to not create the template if system_probe is not enabled and the file doesn't already exists.

Finally, the rspec test fail since we are testing the `system-probe` recipe independently the `dd-agent` recipe, and the `system-probe` template notifies the `datadog-agent` which doesn't exist in this context.
So I refactored the test a bit and introduced a mock `datadog-agent` service into the run context.

### Test plan
1. Run kitchen test for `dd-agent`
The first time the recipe is converged, it should not create the system probe template.
The second time the recipe is converged, it should not update any resource.

2. Run kitchen test for `system-probe`
The first time the recipe is converged, it should create the system probe template and start the service.
The second time the recipe is converged, it should not update any resource.
Disabling `system-probe` in `kitchen.yml` and re-converging, the `system-probe` recipe should update the template file to set `enabled: false` and stop the service.
Re-converging once more after that, no resource should be updated.

3. Set the agent version to 6.11 or less in `kitchen.yml`.
Upon convergence the `system-probe` recipe should not be included.

4. Set the agent version to 6.12+, 7.16, (or any 7.x version) in `kitchen.yml`.
Upon convergence the `system-probe` recipe should be included.

### DIFF
--- a/libraries/recipe_helpers.rb
+++ b/libraries/recipe_helpers.rb
@@ -52,6 +52,20 @@ class Chef
         7
       end
 
+      def agent_minor_version(node)
+        agent_version = agent_version(node)
+        unless agent_version.nil?
+          match = agent_version.match(/([0-9]+:)?([0-9]+)\.([0-9]+)\.([0-9]+)([^-\s]+)?(?:-([0-9]+))?/)
+          if match.nil?
+            Chef::Log.warn "Couldn't infer agent_minor_version from agent_version '#{agent_version}'"
+          else
+            _epoch, _major, minor, _patch, _suffix, _release = match.captures
+            return minor.to_i
+          end
+        end
+        nil
+      end
+
       def agent_flavor(node)
         # user-specified values
         agent_flavor = node['datadog']['agent_flavor']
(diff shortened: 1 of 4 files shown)


albertvaka (commented):
I think this will fail in case someone is installing a very old version of the Agent that does't ship the system-probe binary. Maybe the `service` resource could have an `if` clause that checks the binary is there?

albertvaka:
Good point. I don't know the version, you will have to do a bit of archeology ⚒️

albertvaka (approved):

PR #737, opened 2020-09-02
Release version 4.5.0

julien-lebot:
### What does this PR do?

It releases version 4.5.0


albertvaka (approved):
I would remove 723, 726 and 733 since they have no user-facing impact. But we can also be exhaustive and link everything. Apart from that LGTM.

PR #741, opened 2020-10-02
[circlci] Add tests for Chef 16.5.77

truthbk:


albertvaka:
I've refactored a bit the tests, because it wasn't obvious what "full-suite" meant (it just called kitchen docker tests, so there was no point on creating a copy of it for chef 16 -> instead a new entry had to be added to `kitchen.docker.yaml`)

truthbk (commented):
Nice! 👍

PR #743, opened 2020-10-05
Stricter linter checks

albertvaka:
Update rubocop and foodcritic and fix all the new linting issues.


Rakefile
```diff
@@ -25,7 +25,13 @@ namespace :style do
 
   desc 'Run Chef style checks'
   FoodCritic::Rake::LintTask.new(:chef) do |t|
-    t.options = { fail_tags: ['correctness'] }
+    t.options = {
+      fail_tags: ['correctness'],
+      tags: [
+        '~FC048', # Disables: Prefer shell_out helper method to shelling out with ruby
```
mx-psi, line 31: Why do we disable this one? I assume `FC121` is disabled because we want to support Chef <14 but I am not sure about `FC048`.

libraries/recipe_helpers.rb
```diff
@@ -98,6 +98,25 @@ def cookbook_version(run_context)
         run_context.cookbook_collection['datadog'].version
       end
 
+      def upstart_platform?(node)
+        agent_major_version(node) > 5 &&
+          (((node['platform'] == 'amazon' || node['platform_family'] == 'amazon') && node['platform_version'].to_i != 2) ||
+           (node['platform'] == 'ubuntu' && node['platform_version'].to_f < 15.04) || # chef <11.14 doesn't use the correct service provider
+          (node['platform'] != 'amazon' && node['platform_family'] == 'rhel' && node['platform_version'].to_i < 7))
```
mx-psi, line 105: Why do we have to use `platform_family?` elsewhere but we use `node['platform_family']` here?

recipes/dogstatsd-ruby.rb
```diff
@@ -1,8 +1,8 @@
 #
-# Cookbook Name:: datadog
+# Cookbook:: datadog
 # Recipe:: dogstatsd-ruby
 #
-# Copyright 2013-2105, Datadog
+# Copyright:: 2013-2105, Datadog
```
mx-psi, line 5: While I am happy about the optimism in thinking that this Chef's cookbook will be used for ~80 more years I think we should be more humble 😄

```suggestion
# Copyright:: 2013-2015, Datadog
```
I don't know if we should maybe bump everything to 2020 (in a different PR)?

recipes/repository.rb
```diff
@@ -29,7 +28,7 @@
 
 case node['platform_family']
 when 'debian'
-  include_recipe 'apt'
+  apt_update 'update'
```
mx-psi, line 31: Docs say you can do
```suggestion
  apt_update
```
which sounds simpler to me

albertvaka: The rspec test needs to check for a name, and it doesn't pass if it's not explicitly named: The docs say 'update' is the default name but it's not true at least on the version of chef I've tried.

recipes/dogstatsd-ruby.rb
albertvaka: It's a bit of a hassle to update the copyright every year in all repos, and I'm not sure how important it is from a legal point of view. We should ask.

libraries/recipe_helpers.rb
albertvaka: For some reason, the method is not defined from the `recipe_helpers.rb`. I didn't investigate why.

Rakefile
albertvaka: I was just being lazy, I've fixed the code and re-enabled the check.

recipes/dogstatsd-ruby.rb
mx-psi: 👍 OpenTelemetry for example removed it a while back so maybe it's not needed

libraries/recipe_helpers.rb
mx-psi: 🤔 Okay, it should work the same way so we can leave like that I guess

Rakefile
mx-psi: Looks like it is not available on Chef <14 either 😞

mx-psi (approved):
LGTM, you can revert the last commit since it looks like it won't work with Chef 13

PR #744, opened 2020-10-06
Fix syntax issues in README examples

jimdaga:
## Changes:
* When using this doc as an example of writing my own code I ran across syntax errors when copying examples direct from this repo. This Pull request fixes the examples to use valid syntax.


albertvaka:
Ooops, good catch there! Thanks for the PR 🙇

PR #746, opened 2020-10-08
Increase Windows MSI install timeout from 900 seconds to 1200 seconds

mikezhu-dd:
The current timeout is not long enough in QA environment. Changing it to 1200 seconds.


mx-psi (approved):

olivielpeau (approved):
🙇

PR #749, opened 2020-10-13
Explicitly require yaml

### LINKED ISSUE #747: Run fails with latest Chef client: undefined method `to_yaml' for #<Hash>
After updating to Chef 16.5.17, I am getting the following error:

```
Chef::Mixin::Template::TemplateError (undefined method `to_yaml' for #<Hash:0x000055f965b17830>) on line #178:

176: # Generated by Chef, local modifications will be overwritten
177:
178: <%= JSON.parse(agent_config.to_json).to_yaml %>

 …/cache/cookbooks/datadog/templates/default/datadog.yaml.erb:178:in `block in evaluate'
```

I assume this comes from this change: https://github.com/chef/chef/pull/10383/files#diff-56181b7867c876de4e63eff4b4e06504

The `yaml` gem is no longer loaded all the time, but only when explicitly used, so now there may not be a `Hash#to_yaml` method defined as the gem has not yet been loaded.

Requiring this gem from an early-loaded recipe allows Chef to run without errors :
```ruby
require "yaml"
```

albertvaka:
Chef 16.5.17 no longer does this for us

Should fix #747

### DIFF
--- a/recipes/dd-agent.rb
+++ b/recipes/dd-agent.rb
@@ -17,6 +17,8 @@
 # limitations under the License.
 #
 
+require 'yaml' # Our erb templates need this
+
 if node['datadog'].include?('agent6')
   Chef::Log.warn('The boolean "agent6" is no longer used by this cookbook since version 4.0.0. Use "agent_major_version" instead. To keep the previous behaviour pin a 3.x version of the cookbook.')
 end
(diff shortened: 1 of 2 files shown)


julien-lebot (approved):
LGTM

PR #762, opened 2020-12-09
Take site option into account in chef handler config

KSerrania:
### What does this PR do?

When passing config to chef-handler, take into account the `site` option.

### Motivation

Avoid cases where, with `node['datadog']['site']` set, the Agents report to a given backend, but the chef handler stills sends reports to the default backend (which most likely fails due to a wrong API / APP key for this backend).

### DIFF
--- a/recipes/dd-handler.rb
+++ b/recipes/dd-handler.rb
@@ -65,9 +65,14 @@ chef_handler 'Chef::Handler::Datadog' do
   def handler_config
     extra_config = node['datadog']['handler_extra_config'].reject { |_, v| v.nil? }
 
-    # Since Agent 6 supports node['datadog']['url'] = nil, we need to fallback
-    # on a default value here.
+    # Since Agent 6 supports node['datadog']['url'] = nil and node['datadog']['site'] = nil,
+    # we need to fallback on a default value here.
+
+    # Passing the site option directly to the handler is only supported since v0.12.0 of chef-handler-datadog.
+    # To preserve compatibility with older versions, the site option is therefore
+    # resolved here, and only the URL is passed.
     dd_url = 'https://app.datadoghq.com'
+    dd_url = 'https://app.' + node['datadog']['site'] unless node['datadog']['site'].nil?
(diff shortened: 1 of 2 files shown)


recipes/dd-handler.rb
```diff
@@ -65,9 +65,14 @@ def extra_endpoints
...
-    # Since Agent 6 supports node['datadog']['url'] = nil, we need to fallback
-    # on a default value here.
+    # Since Agent 6 supports node['datadog']['url'] = nil and node['datadog']['site'] = nil,
+    # we need to fallback on a default value here.
+
+    # Passing the site option directly to the handler is only supported since v0.12.0 of chef-handler-datadog.
+    # To preserve compatibility with older versions, the site option is therefore
+    # resolved here, and only the URL is passed.
     dd_url = 'https://app.datadoghq.com'
+    dd_url = 'https://app.' + node['datadog']['site'] unless node['datadog']['site'].nil?
```
julien-lebot, line 75: Should we document it somewhere that `site` has precedence over the `url` ?
Not likely to happen, but what if both `site` and `url` are set ?

albertvaka: If you set both, the Agent gives a Warning on startup, so users should know this is incorrect. Since this uses the settings that are set for the Agent, I think it's okay to not warn here (as the Agent will warn already).

albertvaka (approved):

recipes/dd-handler.rb
KSerrania: > Should we document it somewhere that site has precedence over the url ?
> Not likely to happen, but what if both site and url are set ?

If both are set, the url takes precedence (same as what we do for the config): https://github.com/DataDog/chef-datadog/blob/252589be4b372ab759cf847140f97457ccef2f97/recipes/_agent6_config.rb#L26-L30

This behavior is already documented in the attributes configuration: https://github.com/DataDog/chef-datadog/blob/252589be4b372ab759cf847140f97457ccef2f97/attributes/default.rb#L96-L97

julien-lebot (approved):

PR #764, opened 2020-12-10
Make Key Changes Work With Chef lower 13.4

nilskuehme:
Unfortunately I get the error that Chef in apt_repository cannot read an array in the field "keys".The support of the array works only from chef 13.4. The cookbook is supported from chef 12.7 on according to metadata.rb. In order not to make a jump from 12.7 to 13.4 I have implemented that change. This change would help us a lot.

>         65:    retries = node['datadog']['aptrepo_retries']
>         66:    keyserver = node['datadog']['aptrepo_use_backup_keyserver'] ? node['datadog']['aptrepo_backup_keyserver'] : node['datadog']['aptrepo_keyserver']
>         67:    # Add APT repositories
>         68:    # if Chef::VERSION.to_i >= 13.4
>         69:    apt_repository 'datadog' do
>         70:      keyserver keyserver
>         71>>     key apt_gpg_keys
>         72:      uri node['datadog']['aptrepo']
>         73:      distribution node['datadog']['aptrepo_dist']
>         74:      components components
>         75:      action :add
>         76:      retries retries
>         77:    end
>         78:    # else
>         79:    #   # Works as long as both key entries serves the same content
>         80:    #   apt_gpg_keys.each do |apt_gpg_key|

### DIFF
--- a/recipes/repository.rb
+++ b/recipes/repository.rb
@@ -65,6 +65,7 @@ when 'debian'
   retries = node['datadog']['aptrepo_retries']
   keyserver = node['datadog']['aptrepo_use_backup_keyserver'] ? node['datadog']['aptrepo_backup_keyserver'] : node['datadog']['aptrepo_keyserver']
   # Add APT repositories
+  if Chef::VERSION.to_i >= 13.4
   apt_repository 'datadog' do
     keyserver keyserver
     key apt_gpg_keys
@@ -73,6 +74,20 @@ when 'debian'
     components components
     action :add
     retries retries
+  end
+  else
+    # Works as long as both key entries serves the same content
+    apt_gpg_keys.each do |apt_gpg_key|
+      apt_repository 'datadog' do
+        keyserver keyserver
+        key apt_gpg_key
+        uri node['datadog']['aptrepo']
+        distribution node['datadog']['aptrepo_dist']
+        components components
+        action :add
+        retries retries
+      end
+    end
   end
 
   # Previous versions of the cookbook could create these repo files, make sure we remove it now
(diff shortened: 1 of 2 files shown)


albertvaka:
Hi Nils! 

First of all, thanks a lot for your help with all these changes 😄

It seems that with your change only one key gets added on pre-13.4 according to the tests. Probably the resources created in a loop just override one another and only the last one remains.

If nothing works, we can revert the code to the state where  it only adds one key and add the other key 'manually' like in here:
https://github.com/DataDog/chef-datadog/blob/028aa797c65df1affccf2058e4a9fdb1c7e98d52/recipes/repository.rb#L28-L32

It's uglier, but it works on all chef versions 🤷 

What do you think?

nilskuehme:
Hi @albertvaka ,

that sounds like a good solution to me. So why not...but i would like to use that key for the apt_repository ressource and this one D75CEA17048B9ACBF186794B32637D44F14F620E for the execute ressource. Is this something you want to do?

albertvaka:
It shouldn't matter which key appears where, I don't have a strong opinion either way. I can implement the change later today, unless you wanna do it :)

nilskuehme:
I was experiencing issues while trying to just use one key. Using D75C... packages couldn't be found during the chef-client run. Yeah let me try.

KSerrania:
> I was experiencing issues while trying to just use one key. Using D75C... packages couldn't be found during the chef-client run. Yeah let me try.

That's expected with the current state of the repo. Key `D75CEA17048B9ACBF186794B32637D44F14F620E` is not used yet (but will be in the near future). Currently, key `A2923DFF56EDA6E76E55E492D3A80E30382E94DE` is needed to access the repository.

The reason why we want to trust both keys is that we'll switch our signing key soon, and we want people to trust the new key before we do said switch.

This is why we want something like:
```
apt_repository 'datadog' do
  ...
  key 'A2923DFF56EDA6E76E55E492D3A80E30382E94DE'
  ...
end

execute 'apt-key import key D75CEA17048B9ACBF186794B32637D44F14F620E' do 
   command 'apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 D75CEA17048B9ACBF186794B32637D44F14F620E' 
   not_if 'apt-key list | grep F14F620E' 
 end
```

albertvaka:
New PR: https://github.com/DataDog/chef-datadog/pull/765

PR #765, opened 2020-12-10
Add multiple apt keys in a way that's compatible with chef < 13.4

albertvaka:
Chef older than 13.4 doesn't support passing an array in the `key` field of `apt_repository`. Add a single key there an manually execute `apt-key` to add any additional keys.

### DIFF
--- a/recipes/repository.rb
+++ b/recipes/repository.rb
@@ -63,6 +63,14 @@ when 'debian'
     Chef::Log.error("agent_major_version '#{agent_major_version}' not supported.")
   end
 
+  other_apt_gpg_keys.each do |key|
+    key_short = key[-8..-1] # last 8 chars, since some versions of apt-key list add dashes in between key parts
+    execute "apt-key import key #{key_short}" do
+      command "apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 #{key}"
+      not_if "apt-key list | grep #{key_short}"
+    end
+  end
+
   retries = node['datadog']['aptrepo_retries']
   keyserver = node['datadog']['aptrepo_use_backup_keyserver'] ? node['datadog']['aptrepo_backup_keyserver'] : node['datadog']['aptrepo_keyserver']
   # Add APT repositories
@@ -74,14 +82,6 @@ when 'debian'
     components components
     action :add
     retries retries
-  end
-
-  other_apt_gpg_keys.each do |key|
-    key_short = key[-8..-1] # last 8 chars, since some versions of apt-key list add dashes in between key parts
-    execute "apt-key import key #{key_short}" do
-      command "apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 #{key}"
-      not_if "apt-key list | grep #{key_short}"
-    end
   end
 
   # Previous versions of the cookbook could create these repo files, make sure we remove it now
--- a/recipes/repository.rb
+++ b/recipes/repository.rb
@@ -28,7 +28,8 @@ agent_major_version = Chef::Datadog.agent_major_version(node)
 
 # A2923DFF56EDA6E76E55E492D3A80E30382E94DE expires in 2022
 # D75CEA17048B9ACBF186794B32637D44F14F620E expires in 2032
-apt_gpg_keys = ['A2923DFF56EDA6E76E55E492D3A80E30382E94DE', 'D75CEA17048B9ACBF186794B32637D44F14F620E']
+apt_gpg_key = 'D75CEA17048B9ACBF186794B32637D44F14F620E'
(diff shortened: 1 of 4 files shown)


recipes/repository.rb
```diff
@@ -67,14 +68,22 @@
...
   end
 
+  other_apt_gpg_keys.each do |key|
+    key_short = key[-8..-1] # last 8 chars, since some versions of apt-key list add dashes in between key parts
+    execute "apt-key import key #{key_short}" do
+      command "apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 #{key}"
+      not_if "apt-key list | grep #{key_short}"
+    end
+  end
+
```
KSerrania, line 86: I think you need to put this loop before calling `apt_repository`. As it stands, what chef does is:
- install D75CEA17048B9ACBF186794B32637D44F14F620E (future) key
- apt-get update
- install A2923DFF56EDA6E76E55E492D3A80E30382E94DE (current) key
- apt-get install datadog-agent

Which fails, because `apt-get update` couldn't authenticate our repository (since we didn't trust the key yet).

cf. the results of kitchen tests:
```
         * apt_repository[datadog] action add
           * execute[install-key D75CEA17048B9ACBF186794B32637D44F14F620E] action run
             - execute apt-key adv --no-tty --recv --keyserver hkp://keyserver.ubuntu.com:80 D75CEA17048B9ACBF186794B32637D44F14F620E
           * execute[apt-cache gencaches] action run
             - execute ["apt-cache", "gencaches"]
           * execute[apt-cache gencaches] action nothing (skipped due to action :nothing)
           * apt_update[datadog] action nothing (skipped due to action :nothing)
           * file[/etc/apt/sources.list.d/datadog.list] action create
             - create new file /etc/apt/sources.list.d/datadog.list
             - update content in file /etc/apt/sources.list.d/datadog.list from none to de9651
             --- /etc/apt/sources.list.d/datadog.list	2020-12-10 20:25:46.163306738 +0000
             +++ /etc/apt/sources.list.d/.chef-datadog20201210-93-1npkuv1.list	2020-12-10 20:25:46.163306738 +0000
             @@ -1 +1,2 @@
             +deb      "http://apt.datadoghq.com" stable main
             - change mode from '' to '0644'
             - change owner from '' to 'root'
             - change group from '' to 'root'
           * execute[apt-cache gencaches] action run
             - execute ["apt-cache", "gencaches"]
           * apt_update[datadog] action update
             - force update new lists of packages
             * directory[/var/lib/apt/periodic] action create (up to date)
             * directory[/etc/apt/apt.conf.d] action create (up to date)
             * file[/etc/apt/apt.conf.d/15update-stamp] action create_if_missing (up to date)
             * execute[apt-get -q update] action run
        - execute ["apt-get", "-q", "update"]
           
         
         * execute[apt-key import key 382E94DE] action run
           - execute apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 A2923DFF56EDA6E76E55E492D3A80E30382E94DE
         * apt_repository[datadog-beta] action remove (up to date)
         * apt_repository[datadog_apt_A2923DFF56EDA6E76E55E492D3A80E30382E94DE] action remove (up to date)
         * apt_repository[datadog_apt_D75CEA17048B9ACBF186794B32637D44F14F620E] action remove (up to date)
       Recipe: datadog::_install-linux
         * apt_package[datadog-agent] action install
           
           Error executing action `install` on resource 'apt_package[datadog-agent]'
           
           Mixlib::ShellOut::ShellCommandFailed
           ------------------------------------
           Expected process to exit with [0], but received '100'
           ---- Begin output of ["apt-get", "-q", "-y", "-o", "Dpkg::Options::=--force-confdef", "-o", "Dpkg::Options::=--force-confold", "install", "datadog-agent=1:5.32.8-1"] ----
           STDOUT: Reading package lists...
           Building dependency tree...
           Reading state information...
           The following NEW packages will be installed:
             datadog-agent
           0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
           Need to get 101 MB of archives.
           After this operation, 338 MB of additional disk space will be used.
           WARNING: The following packages cannot be authenticated!
             datadog-agent
           STDERR: E: There are problems and -y was used without --force-yes
           ---- End output of ["apt-get", "-q", "-y", "-o", "Dpkg::Options::=--force-confdef", "-o", "Dpkg::Options::=--force-confold", "install", "datadog-agent=1:5.32.8-1"] ----
           Ran ["apt-get", "-q", "-y", "-o", "Dpkg::Options::=--force-confdef", "-o", "Dpkg::Options::=--force-confold", "install", "datadog-agent=1:5.32.8-1"] returned 100
```

albertvaka: Totally correct. Fixed.

KSerrania (approved):

PR #772, opened 2021-01-20
Fix yum_package[gnupg] resource being declared more than once

cdonadeo:
On RHEL OSes, the datadog::repository recipe declared yum_package[gnupg] once for each GPG key causing deprecation warnings on Chef 12.

I was unable to get rspec to run cleanly before my change, but no new failures were reported after my change.  I was able to run the `dd-agent-centos-71-*` kitchen tests, except for `dd-agent-centos-71-1272` which fails for unrelated reasons (version dependency issues between the `apt` cookbook and Chef 12).

### DIFF
--- a/recipes/repository.rb
+++ b/recipes/repository.rb
@@ -98,15 +98,15 @@ when 'debian'
   end
 
 when 'rhel', 'fedora', 'amazon'
+  # gnupg is required to check the downloaded key's fingerprint
+  package 'gnupg' do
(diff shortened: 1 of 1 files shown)


albertvaka:
Good catch, it doesn't make sense to do this inside the loop. Thanks a lot for the patch! 

And the failing tests seem unrelated indeed, I'll take a look 👍

Issue #775, opened 2021-02-09
Support custom MBean collection


EladLeev:
Hi,
When using the Kafka integration, you can determine what version of Kafka are you running, and the template will [generate](https://github.com/DataDog/chef-datadog/blob/master/templates/default/kafka.yaml.erb) a list of MBeans to collect based on it.

This is great, but Kafka offers [much more](https://gist.github.com/thbkrkr/77c8f6f9a301d7b16555726793af8301) metrics than the sample ones in the template.
It could be great to have the ability to better control the template, and provide another subset of MBeans to collect.

Thanks

lchayoun:
Adding to that, the [example](https://github.com/DataDog/integrations-core/blob/master/kafka/datadog_checks/kafka/data/conf.yaml.example) configuration for the kafka integration has much more options that what is available through the [template](https://github.com/DataDog/chef-datadog/blob/master/templates/default/kafka.yaml.erb).

It would be great to add a way to control all of those options.

albertvaka:
Thanks for reporting this! Indeed, the kafka config template provides some pre-defined metrics that we should probably remove and leave up to the user to configure. Unfortunately this is most likely a breaking change, so we might need to hold that change until the next major version. However, in the meantime, note you can set any check config you want by using the `datadog_monitor` resource, which simply takes a dictionary with the config you want to set. See the instructions here: https://github.com/DataDog/chef-datadog#datadog_monitor

lchayoun:
Hi @albertvaka, thank you for the quick response.
I think that we can use the `version` (`['datadog']['kafka']['version']`) and in version 3 enable the new behavior.

albertvaka:
@lchayoun I've implemented your suggestion in #776 :)

lchayoun:
LGTM 😃

PR #776, opened 2021-02-11
Kafka recipe: allow setting any conf

albertvaka:
Any arguments passed as `node['datadog']['kafka']['instances']` as well as `node['datadog']['kafka']['conf']` will be written in the config file (as `instances` and `init_config.conf` respectively).

Fixes: #775


julien-lebot (approved):
LGTM

PR #777, opened 2021-02-11
Fix fatal in error handler if ddagentuser_name or ddagentuser_password not defined

albertvaka:
Fixes:
```
FATAL: Chef::Exceptions::ResourceNotFound: Cannot find a resource matching windows_env[DDAGENTUSER_NAME] (did you define it first?)
```
Which crashed the error handling flow and didn't let you see the actual error.

It also reorders the code a bit, sorry if that's confusing to review but it's just moving lines around.


mx-psi (approved):
LGTM!

PR #778, opened 2021-02-12
Feature: allow muting a host during an MSI install

albertvaka:
MSI installs can take a long time, during which the Agent is not running. On upgrades, this could trigger no-data monitors.

This PR adds a new option to mute a host while an MSI install takes place, thus preventing this issue. The host will unmute after either the install completes, the chef run fails, or after the mute expires (in 60 minutes).

This uses the [Datadog API ](https://docs.datadoghq.com/api/latest/hosts/#mute-a-host), which requires an `application_key`.


julien-lebot (approved):
LGTM

PR #780, opened 2021-02-17
Support NPM installation switch

mikezhu-dd:
Tested by enabling the switch and running kitchen converge dd-agent-windows-2012r2-13113,
log into the VM, open C:\Users\vagrant\AppData\Local\Temp\MSI*.log and make sure NPM = true and NPM driver is installed.

### DIFF
--- a/recipes/_install-windows.rb
+++ b/recipes/_install-windows.rb
@@ -82,6 +82,8 @@ else
   dd_agent_installer_basename = "#{dd_agent_installer_prefix}-#{dd_agent_version}"
 end
 
+dd_agent_install_npm = Chef::Datadog.npm_install(node)
+
 temp_file_basename = ::File.join(Chef::Config[:file_cache_path], 'ddagent-cli')
 temp_fix_file = ::File.join(Chef::Config[:file_cache_path], 'fix_6_14.ps1')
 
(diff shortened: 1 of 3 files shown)


recipes/_install-windows.rb
```diff
@@ -102,6 +104,8 @@ def unmute_host(context)
   # custom credentials and use them if that's the case.
   install_options.concat(' DDAGENTUSER_NAME=%DDAGENTUSER_NAME%') if ddagentuser_name
   install_options.concat(' DDAGENTUSER_PASSWORD=%DDAGENTUSER_PASSWORD%') if ddagentuser_password
+
+  install_options.concat(" NPM=#{dd_agent_install_npm}") if dd_agent_install_npm
```
albertvaka, line 108:
```suggestion
  install_options.concat(" NPM=#{dd_agent_install_npm}") unless dd_agent_install_npm.nil?
```
Nit: Checking for null explicitly. This way if someone sets `windows_npm_install` to `false` we will still pass it as `false` instead of not passing it. Right now this won't do much since `false` is the default, but that might change in the future.

albertvaka (approved):
Left one comment but LGTM

albertvaka (approved):

recipes/_install-windows.rb
```diff
@@ -102,6 +104,8 @@ def unmute_host(context)
   # custom credentials and use them if that's the case.
   install_options.concat(' DDAGENTUSER_NAME=%DDAGENTUSER_NAME%') if ddagentuser_name
   install_options.concat(' DDAGENTUSER_PASSWORD=%DDAGENTUSER_PASSWORD%') if ddagentuser_password
+
+  install_options.concat(" NPM=#{dd_agent_install_npm}") unless dd_agent_install_npm.nil?
```
albertvaka, line 108:
```suggestion
  install_options.concat(" NPM=#{dd_agent_install_npm}") if dd_agent_install_npm
```
Actually this was correct before: We don't support passing `NPM=false` (it gets treated the same way as passing `true`)

PR #783, opened 2021-03-08
Lint README for Documentation style

ruthnaebeck:
### What does this PR do?
Lint README for Documentation style guidelines

### Motivation
Docs hackathon

### Preview
N/A

### Additional Notes
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,7 +303,7 @@ There are many [recipes][7] to assist you with deploying Agent integration confi
 
 ## Resources
 
-### datadog_monitor
+### Datadog monitor
 
 Use the `datadog_monitor` resource for enabling Agent integrations without a recipe.
 
@@ -352,7 +352,7 @@ end
 
 See the [Datadog integration Chef recipes][7] for additional examples.
 
-### datadog_integration
+### Datadog integration
 
 To install a specific version of a Datadog integration, use the `datadog_integration` resource.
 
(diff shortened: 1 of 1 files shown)


julien-lebot (approved):

cswatt (approved):

README.md
```diff
@@ -303,7 +303,7 @@ There are many [recipes][7] to assist you with deploying Agent integration confi
 
 ## Resources
 
-### datadog_monitor
+### Datadog monitor
```
albertvaka, line 306: This is documenting the `datadog_monitor` resource (ie: class-like thing you can instantiate in chef) in our chef cookbook. By renaming it this could be confused with Datadog monitors (which are a completely different thing, unfortunately). I like the explicit resource name better as a title here, but if you don't want to mention code bits in the title, then this shoul be something like "Setting up integrations without explicit recipes".

README.md
```diff
@@ -352,7 +352,7 @@ end
 
 See the [Datadog integration Chef recipes][7] for additional examples.
 
-### datadog_integration
+### Datadog integration
```
albertvaka, line 355: Same here, `datadog_integration` is a thing our cookbook defines. Alternative title: "Install or upgrade a specific version of an integration"

albertvaka (changes requested):

albertvaka (approved):

PR #789, opened 2021-04-26
Enable yum repository repo_gpgcheck option for Red Hat platforms by default

bkabrda:
I'm wondering whether we should document this - there doesn't seem to be any section in README.md that explicitly deals with gpg keys, repository URLs etc - I'm assuming because the users never really hit issues with these and so they didn't need documenting?

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -160,6 +160,13 @@ default['datadog']['aptrepo_retries'] = 4
 default['datadog']['aptrepo_use_backup_keyserver'] = false
 default['datadog']['aptrepo_keyserver'] = 'hkp://keyserver.ubuntu.com:80'
 default['datadog']['aptrepo_backup_keyserver'] = 'hkp://pool.sks-keyservers.net:80'
+# repo_gpgcheck on RHEL 5 requires additional packages (e.g. pygpgme) that might not
+# always be installed/available in enabled repositories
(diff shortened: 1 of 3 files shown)


attributes/default.rb
```diff
@@ -160,6 +160,13 @@
 default['datadog']['aptrepo_use_backup_keyserver'] = false
 default['datadog']['aptrepo_keyserver'] = 'hkp://keyserver.ubuntu.com:80'
 default['datadog']['aptrepo_backup_keyserver'] = 'hkp://pool.sks-keyservers.net:80'
+# repo_gpgcheck on RHEL 5 requires additional packages (e.g. pygpgme) that might not
+# always be installed/available in enabled repositories
+default['datadog']['yumrepo_repo_gpgcheck'] = if platform_family?('rhel') && node['platform_version'].to_i < 6
```
albertvaka, line 165: If a user has defined their own `yumrepo` URL pointing to a custom repo which isn't signed, this would be a breaking change for them. We can check that and default to false in that case. For example, we could make `yumrepo_repo_gpgcheck` default to `nil`, and then at runtime calculate the actual default (`true` or `false`) based on:
- the existing logic for rhel5.
- whether `yumrepo` is `nil` (default repo) or not (custom repo).
Wdyt?

bkabrda: Good point. I'll fix this.

albertvaka (approved):
👍  LGTM

PR #796, opened 2021-05-14
 Improvements for APT keys management

bkabrda:
* By default, get keys from keys.datadoghq.com, not Ubuntu keyserver
* Always add the DATADOG_APT_KEY_CURRENT.public key (contains key used to sign current repodata)
* Add 'signed-by' option to all sources list lines
* On Debian >= 9 and Ubuntu >= 16, only add keys to /usr/share/keyrings/datadog-archive-keyring.gpg
* On older systems, also add the same keyring to /etc/apt/trusted.gpg.d

### DIFF
--- a/recipes/repository.rb
+++ b/recipes/repository.rb
@@ -26,7 +26,7 @@ yum_a5_architecture_map.default = 'x86_64'
 
 agent_major_version = Chef::Datadog.agent_major_version(node)
 
-# DATADOG_APT_KEY_CURRENT always contains the key that is used to sing repodata and latest packages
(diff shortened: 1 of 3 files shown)


recipes/repository.rb
```diff
@@ -26,10 +26,18 @@
 
 agent_major_version = Chef::Datadog.agent_major_version(node)
 
+# DATADOG_APT_KEY_CURRENT always contains the key that is used to sing repodata and latest packages
```
mx-psi, line 29: 🧑‍🎤 🎙️ 
```suggestion
# DATADOG_APT_KEY_CURRENT always contains the key that is used to sign repodata and latest packages
```

mx-psi (approved):

PR #799, opened 2021-05-21
Ensure NPM is not turned off by default when system probe is enabled

### LINKED ISSUE #798: Breaking change introduced in version 4.10.0 - network monitoring
In this PR, the network monitoring attribute was changed, breaking existing configurations:
https://github.com/DataDog/chef-datadog/pull/784/files

bkabrda:
In order to preserve the pre-4.10.0 behavior, we need to enable network monitoring whenever system probe is enabled, unless this is overriden explicitly.

Fixes #798.


KSerrania (commented):
Would it be possible to not set the `network_config.enabled` option at all if it's not explicitly set by the user?

The main issue with the current solution is that chef will enforce `system_probe_config.enabled: true` => `network_config.enabled: true` in the config file, which gives the same behavior as the current system-probe, but may not be true in future versions of system-probe.

bkabrda:
@KSerrania ok, I think it's done now. Basically this is a much simpler version of the PR, which basically just recognizes `nil` as a special value for `network_enabled` and doesn't render `network_config` at all when this value is used. This works thanks to the fact that setting `network_enabled: true` will win over `system_probe.enabled: false`, so even if users just turn on NPM and leave the system probe "disabled" (which is the default value), NPM will still work.

KSerrania (approved):

Issue #805, opened 2021-07-26
integration resource should notify a restart of the agent after installing a new version


lchayoun:
As documented here: https://docs.datadoghq.com/agent/guide/integration-management/#install
"After upgrading, restart your Agent to begin using the newly installed integration"

The issues is, that after upgrading an agent version, all the integration conf files are the same so the agent is not restarted after installing the new agent version and the integration on it.

Same issue occurs when running the cookbook on a new instance, the agent default version is running instead of the newly installed version

lchayoun:
@albertvaka Any schedule for a new release?

albertvaka:
Hi @lchayoun! We should release a new version soon, hopefully this week :)

PR #806, opened 2021-07-30
#805 integration resource should notify a restart of the agent after installing a new version

lchayoun:
Adding `notifies`


albertvaka (approved):
Good catch! Thanks a lot for the patch :)

PR #819, opened 2021-09-13
(DOCS-2444) Remove broken link

jtappa:
There used to be a section that contained this sentence, and that's all that was in it. The anchor was broken and being flagged by docs' pipelines as not existing, so in this case I just moved the content that used to be there into this sentence and removed the anchor link.


kayayarai (approved):

mx-psi (approved):
Thanks!

PR #828, opened 2021-11-03
Update datadog agent 6 yaml to include dogstatsd_port by default

brentm5:
This change allows the dogstatsd_port attribute to be respected [similar to version 5](https://github.com/DataDog/chef-datadog/blob/main/templates/default/datadog.conf.erb#L110) without having to use the extra_config attribute. This allows the [existing attribute to be respected
](https://github.com/DataDog/chef-datadog/blob/main/attributes/default.rb#L280)

Currently with agent 6 the only way to override this is to set the following
```
default['datadog']['extra_config']['dogstatsd_port'] = 8125
```

### DIFF
--- a/spec/dd-agent_spec.rb
+++ b/spec/dd-agent_spec.rb
@@ -1066,6 +1066,7 @@ describe 'datadog::dd-agent' do
             log_file: "/var/log/datadog/agent.log"
             log_level: "INFO"
             dogstatsd_non_local_traffic: false
+            dogstatsd_port: 8125
             apm_config:
               apm_non_local_traffic: false
             process_config:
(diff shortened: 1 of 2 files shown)


albertvaka:
This makes sense, thanks a lot for your contribution! 👍

lchayoun:
I think this should have been marked as a breaking change in the changelog or to make this change backward compatible.
It caused `default['datadog']['extra_config']['dogstatsd_port']` to be ignored and only `default[:datadog][:dogstatsd_port]` is changing the port.

lchayoun:
The reason it made a breaking change is a really old commit that added the default port: https://github.com/DataDog/chef-datadog/commit/fdfed9aa541125a8c71e7cdea1cc6ea5adf786fd

albertvaka:
Ooops, I didn't expect this to be a breaking change. I hope it didn't break your setup @lchayoun 🙏

Maybe we should remove the defaults (or at least the port one) to make it non-breaking.

PR #832, opened 2021-12-24
Use conditional cookbook dependencies

albertvaka:
Should fix the following error in the tests:
```
Chef::Exceptions::CookbookChefVersionMismatch:
  Cookbook 'yum' version '7.2.1' depends on chef version [">= 15.3"], but the running chef version is 14.10.9
```


KSerrania (approved):

albertvaka:
Actually with the current `Gemfile.lock` in the repo the tests don't fail anyway. I'm not sure if we still want this change.

PR #833, opened 2021-12-30
Use the correct socket for system probe on Windows

julien-lebot:


attributes/default.rb
```diff
@@ -369,7 +369,15 @@
 # the NPM module of system probe will still run.
 default['datadog']['system_probe']['enabled'] = false
 # sysprobe_socket defines the unix socket location
-default['datadog']['system_probe']['sysprobe_socket'] = '/opt/datadog-agent/run/sysprobe.sock'
+default['datadog']['system_probe']['sysprobe_socket'] =
+  if platform_family?('windows')
+    # might be //opt/datadog-agent/run/sysprobe.sock
+    # note the //
```
derekwbrown, line 375: the `//` comment should probably be in the `else` clause.

derekwbrown (approved):

attributes/default.rb
julien-lebot: The comment was for myself since I hesitated between `localhost:3333` and `//opt/datadog-agent/run/sysprobe.sock`. I found the `//` syntax [here](https://tomgregory.com/running-docker-in-docker-on-windows/) and thought it might apply here, so I'll just remove the comment.

KSerrania (approved):

PR #838, opened 2022-02-17
Use supported CircleCI image on CI

mx-psi:
See https://circleci.com/docs/2.0/images/linux-vm/14.04-to-20.04-migration


mx-psi:
This did not work :( We are tracking this internally and will fix this, just not on this PR

PR #840, opened 2022-03-04
Kafka Consumer Additions

nilskuehme:
Achieved:

Two options have been added to ```kafka_consumer.rb```:
* monitor_unlisted_consumer_groups
* kafka_consumer_offsets
both work with ```true``` or false. 

monitor_unlisted_consumer_groups:
Because ```monitor_unlisted_consumer_groups``` should only be enabled when no consumer groups are specified it will be written to ```kafka_consumer.d/conf.yaml``` with ```false``` value. Default behavior of the ```conf.yaml``` is that ```monitor_unlisted_consumer_groups``` is set to default when it's not actively enabled with ```true``` but it makes sense to get a chef'd outcome with a like ```monitor_unlisted_consumer_groups: false``` even when it was set to ```true``` when consumer groups were also specified. Like a little reminder instead of not writing it into the config.

kafka_consumer_offsets:
If ```kafka_consumer_offsets``` is set to ```true``` it appears in ```kafka_consumer.d/conf.yaml`` as ```kafka_consumer_offsets: true``` if it's set to ```false``` it will not added to the config as there is no dependency to other attributes.

```kitchen.yaml``` now has Debian 9 and 10 as an test option.

```.bat``` test file has been replaced with serverspec test.


albertvaka (approved):
LGTM, thanks for the patch 🙇

PR #820, opened 2021-09-14
Update to current Chef community conventions

jeffbyrnes:
~Lots of cleanup here, primarily driving towards Chef 17 compatibility.~

Provides:

* Updated support/config files, consistent w/ current Chef community conventions
* Unit tests are performed against latest Chef
* Integration tests are performed against multiple versions of Chef
* Linting is done via `cookstyle`

<details><summary><h3>[Update 2022-08-25]</h3></summary>

Buried in a [comment way down is this, from me:](https://github.com/DataDog/chef-datadog/pull/820#issuecomment-1184835657)

> I looked at the unit test output for https://github.com/DataDog/chef-datadog/pull/852, and it may be that this particular unit test didn’t get run? Quite a few [resources were untouched by that ChefSpec run](https://app.circleci.com/pipelines/github/DataDog/chef-datadog/597/workflows/11a235ec-06a2-433f-9576-e168755d749d/jobs/7625/steps?invite=true#step-107-82).
>
> I ran the unit tests locally against https://github.com/DataDog/chef-datadog/commit/e77a6c3e8cf203ffdabafb8167932d948ceafe74 & a lot of the tests failed so… I think maybe the nature of how the tests are run on main is obfuscating things.
>
> This PR, in part, fixes that & clears up the running of tests, and also ensures the integration tests run against multiple distros, distro versions, Chef versions, and Datadog Agent versions, which the current main branch does not do (despite appearances).

Something to note that this PR brings to bear. I’ve not attempted to _fix_ these test failures, b/c I think that’s out-of-scope for these changes, but we should be mindful all the same.
</details>

<details><summary><h3>[Update 2022-09-13]</h3></summary>

Figured out a bit more about the unit tests while doing a larger cleanup of them (a separate PR will appear for that once this one actually merges).

Check the individual commits to see what sorted that out in the end.
</details>

### [Update 2022-10-12]

* Rebased against `main` branch again
* Added `cookstyle --autofix` lint fixes to make builds totally green

PR #841, opened 2022-03-28
Refactor CircleCI config

jeffbyrnes:
Originally included with #820, this is now broken out into its own PR.

This makes clearer use of the community Chef resources for performing style, unit, & integration tests.

It also more clearly shows which Test Kitchen suites are run, and their individual build status.


jeffbyrnes:
So the lint tests will never pass (lemme know if they should be removed).

The unit tests also will not pass b/c they’re referring to OS versions that Fauxhai no longer provides. See jeffbyrnes/chef-datadog@5a81449f052ab5b69e60795cdcf99e0668a1d73d for details.

The integration tests are failing for reasons unknown; they check that’s failing doesn’t seem to have a good reason to be failing…

PR #853, opened 2022-07-02
[chef] fips-proxy package support

truthbk:
This is barebones package support for a potential `datadog-fips-proxy` package.


attributes/default.rb
```diff
@@ -42,18 +42,27 @@
 default['datadog']['agent_version'] = nil # nil to install latest
 # Agent flavor to install, acceptable values are "datadog-agent", "datadog-iot-agent"
 default['datadog']['agent_flavor'] = 'datadog-agent' # "datadog-agent" to install the datadog-agent package
+# default['datadog']['fips_proxy_version'] = '0.1.0'
```
bkabrda, line 45: Is there a reason why this line is commented out? I think we should do the same thing as we do for agent_version by default, which would be setting it to `nil`.

recipes/dd-fips-proxy.rb
```diff
@@ -0,0 +1,60 @@
...
+if is_windows
+  raise 'Windows currently unsupported for the datadog FIPS proxy.'
+else
+  include_recipe '::_install-fips-proxy-linux'
+end
+
+# Set the Datadog FIPS proxy service enable or disable
+fips_proxy_enable = node['datadog']['fips_proxy_enable'] ? :enable : :disable
+# Set the Datadog FIPS proxy service startup action
+fips_proxy_start = node['datadog']['fips_proxy_start'] ? :start : :stop
```
bkabrda, line 34: Could we have default values in `attributes/defaulr.rb` for `fips_proxy_enable` and `fips_proxy_start`?

recipes/dd-fips-proxy.rb
```diff
@@ -0,0 +1,60 @@
...
+service_provider = Chef::Datadog.service_provider(node)
+
+service_name = 'datadog-fips-proxy'
+
+file '/etc/datadog-fips-proxy/datadog-fips-proxy.cfg' do
+  owner 'root'
+  group 'root'
+  mode 0766
+  content lazy { ::File.open('/etc/datadog-fips-proxy/datadog-fips-proxy.cfg.example').read }
+  not_if { ::File.exist?('/etc/datadog-fips-proxy/datadog-fips-proxy.cfg') }
```
bkabrda, line 46: Is the intention here to not overwrite the config file if it already exists? Most config management tools intentionally ensure that a file has specified contents. I'd prefer if we introduced config variables for what is supposed to go in this file (with the option to using "default" which would use the example file contents).

recipes/dd-fips-proxy.rb
```diff
@@ -0,0 +1,60 @@
...
+fips_proxy_enable = node['datadog']['fips_proxy_enable'] ? :enable : :disable
+# Set the Datadog FIPS proxy service startup action
+fips_proxy_start = node['datadog']['fips_proxy_start'] ? :start : :stop
+
+# Common configuration
+service_provider = Chef::Datadog.service_provider(node)
+
+service_name = 'datadog-fips-proxy'
+
+file '/etc/datadog-fips-proxy/datadog-fips-proxy.cfg' do
```
bkabrda, line 41: A change in this file should notify the fips-proxy service to restart.

bkabrda (changes requested):
:wave: thanks for sending the PR! It mostly looks good, but I left some inline comments to discuss/get addressed.

attributes/default.rb
```diff
@@ -42,18 +42,27 @@
 default['datadog']['agent_version'] = nil # nil to install latest
 # Agent flavor to install, acceptable values are "datadog-agent", "datadog-iot-agent"
 default['datadog']['agent_flavor'] = 'datadog-agent' # "datadog-agent" to install the datadog-agent package
+default['datadog']['fips_proxy_version'] = nil
+default['datadog']['agent_version'] = nil # nil to install latest
```
bkabrda, line 46: This already exists above on line 42, I think this line was added by mistake?

attributes/default.rb
```diff
@@ -42,18 +42,27 @@
...
 default['datadog']['agent_package_action'] = 'install'
+# Allow override with `upgrade` to get latest (Linux only)
+default['datadog']['fips_proxy_package_action'] = 'install'
 
 # Agent package options
 # retries and retry_delay for package download/install
 default['datadog']['agent_package_retries'] = nil
 default['datadog']['agent_package_retry_delay'] = nil
+default['datadog']['fips_proxy_package_retries'] = nil
+default['datadog']['fips_proxy_package_retry_delay'] = nil
```
bkabrda, line 60: I would probably suggest using `agent_package_retries` and `agent_package_retry_delay` for both the Agent and the FIPS proxy. I know that the naming suggests that this would only apply to Agent, but I find it hard to believe that people would make these values different for Agent and FIPS proxy - most of the time, people will have to set both to exactly the same value and I don't think that's a great user experience. WDYT?

metadata.rb
```diff
@@ -3,7 +3,7 @@
 maintainer_email 'package@datadoghq.com'
 license          'Apache-2.0'
 description      'Installs/Configures datadog components'
-version          '4.18.0'
+version          '4.20.0'
```
bkabrda, line 6: Version shouldn't be updated as part of a feature PR.

recipes/dd-fips-proxy.rb
```diff
@@ -0,0 +1,61 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'yaml' # Our erb templates need this
+
+is_windows = platform_family?('windows')
+
+# Install the agent
```
bkabrda, line 24:
```suggestion
# Install the fips-proxy
```

recipes/_install-fips-proxy-linux.rb
```diff
@@ -0,0 +1,76 @@
...
+  else
+    yum_package dd_fips_proxy_package_name do
+      version dd_fips_proxy_version
+      retries package_retries unless package_retries.nil?
+      retry_delay package_retry_delay unless package_retry_delay.nil?
+      action package_action # default is :install
+      allow_downgrade node['datadog']['fips_proxy_allow_downgrade']
+    end
+  end
+when 'suse'
```
bkabrda, line 67: I think we don't have SUSE packages for fips-proxy, meaning that this part should be excluded and we should fail early for SUSE in the dd-fips-proxy recipe (similar to what we do for Windows)

bkabrda (changes requested):
Hi :wave: thanks for addressing the feedback. I think the PR looks pretty solid, but I still left couple comments inline to get addressed.

recipes/dd-fips-proxy.rb
truthbk: Yes, that's the intention. 

The complexity of this file is high, and should not really ever be touched by customers, hence the original decision. The number of potential HAproxy config changes a customer could implement in that file is really quite insane.

bkabrda (approved):

PR #856, opened 2022-08-09
Don't use epoch/release for yum package version on Chef < 14

bkabrda:
The yum logic has been rewritten in Chef 14 and using epoch/release is correct (and necessary), however on Chef < 14 not using epoch/release is correct (and necessary).


amenasria (approved):
LGTM

PR #859, opened 2022-09-21
Add unit types to Systemd integration

nilskuehme:
Initially the Pull Request was created to add the functionality to timer units only. I realized it's the best to add the possibility to add the functionality for all Systemd unit types. 

From my point of view it makes the most sense to let the user select the unit type instead of handing over every string added to node attributes in `systemd.rb`. Therefore an array with all unit types checks if the unit type has been added to node attributes inside the recipe as part of a loop. For a smaller code the decision has been chosen to remove the `s` from the unit type keys inside the hash. It could be a little bit confusing for the end user but in the end it's ways better to have an efficient code in place. 

Node attributes changed from:
```
node.default['datadog']['systemd']['instances'] = [
[...]
    'substate_status_mapping' => [
      'services' => [
        'myservice1' => {
          'running' => 'ok',
          'exited' => 'critical'
        },
[..]
```
to:
```
node.default['datadog']['systemd']['instances'] = [
[...]
    'substate_status_mapping' => [
      'service' => [
        'myservice1' => {
          'running' => 'ok',
          'exited' => 'critical'
        },
[..]
```
Means e.g. the unit type is not called `services` anymore but `service`. 

The array distinguishing between unit types:
```
%w(target service socket device mount automount swap path timer snapshot slice scope busname)
```

UPDATE:
The decision has been chosen to keep the plural version, so the first example has to be used.

### DIFF
--- a/recipes/systemd.rb
+++ b/recipes/systemd.rb
@@ -1,48 +1,48 @@
 include_recipe '::dd-agent'
 
 # Monitor Systemd
-# It works for all unit types like target, service, socket, device, mount, automount, swap, path, timer, snapshot, slice, scope, busname
-# node.default['datadog']['systemd']['instances'] = [
-#   {
-#     'unit_names' => [
-#       'myservice1.service',
-#       'myservice2.service',
-#       'mysocket.socket',
-#       'mytimer.timer'
-#     ],
-#     'substate_status_mapping' => [
-#       'service' => [
-#         'myservice1' => {
-#           'running' => 'ok',
-#           'exited' => 'critical'
-#         },
-#         'myservice2' => {
-#           'plugged' => 'ok',
-#           'mounted' => 'ok',
-#           'running' => 'ok',
-#           'exited' => 'critical',
-#           'stopped' => 'critical'
-#         }
-#       ],
-#       'socket' => [
-#         'mysocket' => {
-#           'running' => 'ok',
-#           'exited' => 'critical'
-#         }
-#       ],
-#       'timer' => [
-#         'mytimer' => {
-#          'running' => 'ok',
-#          'exited' => 'critical'
-#         }
-#       ]
-#     ],
-#     'tags' => [
-#       'mykey1:myvalue1',
-#       'mykey2:myvalue2'
-#     ]
-#   }
-# ]
+# It works for all unit types like targets, services, sockets, devices, mounts, automounts, swaps, paths, timers, snapshots, slices, scopes, busnames
+node.default['datadog']['systemd']['instances'] = [
+  {
+    'unit_names' => [
+      'myservice1.service',
+      'myservice2.service',
+      'mysocket.socket',
+      'mytimer.timer'
+    ],
+    'substate_status_mapping' => [
+      'services' => [
+        'myservice1' => {
+          'running' => 'ok',
+          'exited' => 'critical'
+        },
+        'myservice2' => {
+          'plugged' => 'ok',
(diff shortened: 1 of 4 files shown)


nilskuehme:
Hi @albertvaka, are there any updates? :smiley:

amenasria:
👋 Hey @nilskuehme, sorry for the late reply. Thanks a lot for your PR !

Lot of work right now, our team will review it in the following weeks.

nilskuehme:
Hi @amenasria, yeah I can understand that. Unfortunately we need all 3 changes in our production as soon as possible. Is there any way to take a look on it from you side earlier? That would help us a lot. 

Thank you very much.

templates/default/systemd.yaml.erb
```diff
@@ -9,22 +9,14 @@ instances:
...
-          <% s['sockets'].each do |k| -%>
-            <% k.keys.each do |j| -%>
-      <%= j %>.socket:
-              <% k[j].each do |l,v| %>
+        <% %w(target service socket device mount automount swap path timer snapshot slice scope busname).each do |unit_type| -%>
+          <% if s.key?(unit_type)-%>
+            <% s[unit_type].each do |k| -%>
+              <% k.keys.each do |j| -%>
+      <%= j %>.<%= unit_type %>:
+                <% k[j].each do |l,v| %>
```
julien-lebot, line 17: I prefer the plural version of the unit types, as it's communicates more clearly to the user that multiple instances can be added.
```suggestion
        <% %w(targets services sockets devices mounts automounts swaps paths timers snapshots slices scopes busnames).each do |unit_type| -%>
          <% if s.key?(unit_type)-%>
            <% s[unit_type].each do |k| -%>
              <% k.keys.each do |j| -%>
      <%= j %>.<%= unit_type.chomp('s') %>:
                <% k[j].each do |l,v| %>
```

julien-lebot (commented):
Thank you for this PR, it's good to support all `systemd` unit types; however I still think we should keep the plural version of the `unit_type`. First this will make sure that any existing user of this recipe can still use their code unchanged, and secondly I think the intent is clearer with the plural version and it requires some very minimal code changes to implement.

templates/default/systemd.yaml.erb
nilskuehme: Yeah that make sense. Will commit this suggestion and will adjust the `systemd.rb` example to use the plural version.

nilskuehme:
@julien-lebot are you able to merge the Pull Requests? For me it's not completely clear who is part of the agent-platform.

nilskuehme:
Hi @julien-lebot, no worries. Thank you very much for the clarification :smiley:

recipes/systemd.rb
```diff
@@ -1,41 +1,48 @@
...
+         'exited' => 'critical'
+        }
+      ]
+    ],
+    'tags' => [
+      'mykey1:myvalue1',
+      'mykey2:myvalue2'
+    ]
+  }
+]
```
KSerrania, line 45: I think this example should remain in a comment, to be in line with how other integrations are defined.

KSerrania (commented):
Thanks for the contribution! I left one small comment, otherwise LGTM.

recipes/systemd.rb
nilskuehme: Hi @KSerrania, yeah you are right. I simply forgot to comment that in. Will adjust the PR. Thank you for the hint!

KSerrania (approved):

PR #860, opened 2022-09-23
Add configuration for gearmand integration

nilskuehme:
Achieved:

This PR enables the opportunity to handle the `gearmand` integration via Chef. It's possible to define every parameter described within the integration documentation example file. 

Because all parameters for the `gearmand` integration are optional, the possibility of optional node attributes has been enabled within the configuration deployment, too. Means, if a parameter is not defined, it will be written with default parameters. Well known that the example config file doesn't need that, because these parameters are set by default behind the scenes, the behavior has been set anyway in case of visibility. So the user has the chance to see what he does.

If the user would only add the following to his recipe `node.default['datadog']['gearmand']['instances'] = [{}]`,  the basic standard config would be written as well. If only a view parameters will not be set but others, too - the default settings will be written in case of visibility as well. 

From my point of view this is the most valuable way to handle optional arguments of the DataDog config file via configuration deployment.

### DIFF
--- a/recipes/gearmand.rb
+++ b/recipes/gearmand.rb
@@ -1,0 +1,39 @@
+include_recipe '::dd-agent'
+
+# All attributes are optional.
+# node.default['datadog']['gearmand']['instances'] = [
+#   {
+#     # Defaults to 127.0.0.1 if not set
+#     'server' => '127.0.0.1',
+#     # Defaults to 4730 if not set
+#     'port' => '4730',
+#     'tasks' => [
+#       'TASK_1',
+#       'TASK_2',
+#     ],
+#     'tags' => [
+#       '<KEY_1>:<VALUE_1>',
+#       '<KEY_2>:<VALUE_2>'
+#     ],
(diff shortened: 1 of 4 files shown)


nilskuehme:
Hi @albertvaka, are there any updates? :smiley:

amenasria:
👋 Hey @nilskuehme, sorry for the late reply. Thanks a lot for your PR !

Lot of work right now, our team will review it in the following weeks.

nilskuehme:
Hi @amenasria, yeah I can understand that. Unfortunately we need all 3 changes in our production as soon as possible. Is there any way to take a look on it from you side earlier? That would help us a lot. 

Thank you very much.

julien-lebot (approved):
LGTM thank you for this contribution 🙇 :shipit:

KSerrania (approved):
Thanks for your contribution!

nilskuehme:
Thank you very much guys :rocket:

PR #864, opened 2022-10-28
Add tls_ignore_warning and persist_connections parameters to rabbitmq monitor

bitle:
I have an issue where Datadog Agent is consuming too much CPU when RabbitMQ monitor configuration doesn't have these parameters enabled.

Parameter name reference: https://github.com/DataDog/integrations-core/blob/master/rabbitmq/datadog_checks/rabbitmq/data/conf.yaml.example


AliDatadog:
Closed for inactivity.

PR #865, opened 2022-11-10
Remove source file and datadog-signing-keys package when removing the Agent

### LINKED ISSUE #847: remove_dd_agent recipe does not remove the apt source nor datadog-signing-keys package
We ran into an issue recently where some instances were reporting warnings when trying to perform apt updates:

```
W: An error occurred during the signature verification. The repository is not updated and the previous index files will be used. GPG error: http://apt.datadoghq.com/ stable Release: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 33EE313BAD9589B7
W: Failed to fetch http://apt.datadoghq.com/dists/stable/Release.gpg  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 33EE313BAD9589B7
W: Some index files failed to download. They have been ignored, or old ones used instead.
```

These instances did not in fact have the datadog-agent installed, but it had been installed at some time in the past and was removed, using the `remove_dd_agent` chef recipe.

However it appears that this recipe only purges the datadog-agent package, but leaves behind the source file and the datadog-signing-keys package that are installed by the agent install recipe.

The recipe to remove the datadog agent should remove these files and packages as well as they are installed by this cookbook.

amenasria:
This PR fixes  #847 .

This PR also adds some tests to check that once deleting the Agent, the source file and the `datadog-signing-keys` package are also removed from the host.

### DIFF
--- a/libraries/recipe_helpers.rb
+++ b/libraries/recipe_helpers.rb
@@ -7,6 +7,11 @@ class Chef
         datadog-iot-agent
       ].freeze
 
+      # This method stores a variable that is used across recipes so we needed a place to define it
+      def apt_sources_list_file
+        '/etc/apt/sources.list.d/datadog.list'
+      end
+
       def chef_version_ge?(version)
         Gem::Requirement.new(">= #{version}").satisfied_by?(Gem::Version.new(Chef::VERSION))
       end
--- a/libraries/recipe_helpers.rb
+++ b/libraries/recipe_helpers.rb
@@ -7,7 +7,9 @@ class Chef
         datadog-iot-agent
       ].freeze
 
-      # This method stores a variable that is used across recipes so we needed a place to define it
+      # This method stores a variable that is used across recipes so we needed a place to define it.
+      # Global variables like this are usually set in the attributes/default.rb file for users to edit them.
+      # We don't want this variable to be changed by the user though, hence the place.
       def apt_sources_list_file
         '/etc/apt/sources.list.d/datadog.list'
       end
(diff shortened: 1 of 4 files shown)


recipes/remove-dd-agent.rb
```diff
@@ -1,13 +1,52 @@
...
 case node['os']
 when 'linux'
+  # First remove the datadog-agent package
   package 'datadog-agent' do
     action :purge
   end
+
+  # Then remove the installation files (depending on the OS: sources_list file, datadog-signing-keys package...)
+  case node['platform_family']
+  when 'amazon', 'rhel' # 'rhel' includes redhat, centos, rocky, scientific and almalinux
```
bkabrda, line 31: I think you're missing `fedora` here, this should match https://github.com/DataDog/chef-datadog/blob/main/recipes/repository.rb#L165

bkabrda (changes requested):
Hey :wave: this mostly looks good, but I left some minor comments to address. Thanks!

libraries/recipe_helpers.rb
```diff
@@ -7,6 +7,11 @@ class << self
         datadog-iot-agent
       ].freeze
 
+      # This method stores a variable that is used across recipes so we needed a place to define it
```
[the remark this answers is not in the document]
amenasria: Right, that'd be clearer ! Fixed in [399a428](https://github.com/DataDog/chef-datadog/pull/865/commits/399a42842e3f60b0c92227a2ae59c0d3b6c32e82).

recipes/remove-dd-agent.rb
amenasria: Forgot to put it back, mb ! Fixed in [5c87ee8](https://github.com/DataDog/chef-datadog/pull/865/commits/5c87ee892010d8c3d9c05c6e3bc86113cd1412e7).

PR #870, opened 2022-12-01
Use `chef_kitchen_apt_debian_8` as an image for `Debian-8.11`

amenasria:
The `dd-agent-handler5-debian-811` kitchen instance is failing to build. This is not about the code but about Debian 8 (Jessie) GPG keys that expired. Since they're not going to be renewed we patch the `apt` we're using [a new Docker Image](https://github.com/DataDog/docker-library/pull/88) with the same fix as [this PR](https://github.com/DataDog/datadog-agent-buildimages/pull/319).

This is basically ignoring the KEYEXPIRED errors while still verifying packages signatures.

### DIFF
--- a/kitchen.docker.yml
+++ b/kitchen.docker.yml
@@ -39,6 +39,9 @@ platforms:
   - name: debian-8.11
     driver_config:
       require_chef_omnibus: 14.12
+      # The Debian 8 (Jessie) GPG key expired on Sat Nov 19 2022 21:01:13 GMT+0000.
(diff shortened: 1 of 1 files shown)


KSerrania (changes requested):
The same error is still happening if you check the job logs: https://app.circleci.com/pipelines/github/DataDog/chef-datadog/672/workflows/21944ff1-a621-43d8-b154-56657b472207/jobs/8454

KSerrania (approved):

PR #874, opened 2022-12-16
Actually enforce `apt` and `yum` versions in the `metadata.rb` file

amenasria:
### What does this PR do?

This PR enforces the `apt` and `yum` versions in the `metadata.rb` file.

### Motivation

The `apt` and `yum` dependencies needed versions were explained in the `metadata.rb` file but just as comments:

```ruby
depends    'apt' # Use '< 6.0.0' with Chef < 12.9
depends    'yum', '>= 3.0' # Use '< 5.0' with Chef < 12.14
```

This was also described in the `README.md` file.

This is not really user-friendly neither the best way of doing so since we can just add some logic in the `metadata.rb` file to deal with these cases.

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -22,13 +22,13 @@ issues_url       'https://github.com/DataDog/chef-datadog/issues'
   supports os
 end
 
-if Chef::VERSION.instance_of? String
-  # For Chef < 13.1.33 Chef::VERSION was a String
-  current_chef_version = Gem::Version.new(Chef::VERSION.to_f)
-else
-  # Else Chef::VERSION is already a VersionString
-  current_chef_version = Chef::VERSION
-end
+current_chef_version = if Chef::VERSION.instance_of? String
+    # For Chef < 13.1.33 Chef::VERSION was a String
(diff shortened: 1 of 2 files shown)


urseberry (approved):

metadata.rb
```diff
@@ -27,5 +27,17 @@
...
+# Use '< 6.0.0' with Chef < 12.9
+if Chef::VERSION < Gem::Version.new(12.9)
+  depends 'apt', '< 6.0'
+else
+  depends 'apt'
+end
+
+# Use '< 5.0' with Chef < 12.14
+if Chef::VERSION < Gem::Version.new(12.14)
+  depends 'yum', '< 5.0'
```
bkabrda, line 40: So this should be both >= 3.0 and < 5.0, right?

amenasria: Right ! Nice catch, fixed in [166fd1f](https://github.com/DataDog/chef-datadog/pull/874/commits/166fd1f0d6e36b6d1b5b13681b2638eb0039b760).

chouetz (approved):

PR #877, opened 2023-01-12
Remove old rpm gpg key

amenasria:
### What does this PR do?

This PR removes an formerly used RPM GPG key because it's not used anymore.

### Motivation

It's not used anymore


bkabrda (approved):

amenasria:
Kitchen tests were realised manually 🫠

PR #890, opened 2023-05-30
Add new APT and RPM signing keys

chouetz:
### What does this PR do?
Addition of the new key for signing rotation process

### Motivation
Regular signing key rotation process

Install the agent packages pushed on trial repository

### DIFF
--- a/recipes/repository.rb
+++ b/recipes/repository.rb
@@ -48,9 +48,9 @@ apt_repo_uri = 'https://apt.datadoghq.com'
 # DATADOG_RPM_KEY_FD4BF915.public expires in 2024
 # DATADOG_RPM_KEY_B01082D3.public expires in 2028
 rpm_gpg_keys = [['DATADOG_RPM_KEY_CURRENT.public', 'current', ''],
-                ['DATADOG_RPM_KEY_E09422B3.public', 'e09422b3', 'A4C0 B90D 7443 CF6E 4E8A  A341 F106 8E14 E094 22B3'],
-                ['DATADOG_RPM_KEY_FD4BF915.public', 'fd4bf915', 'C655 9B69 0CA8 82F0 23BD  F3F6 3F4D 1729 FD4B F915']]
-                ['DATADOG_RPM_KEY_B01082D3.public', 'b01082d3', '7408 BFD5 6BC5 BF0C 361A  AAE8 5D88 EEA3 B010 82D3']]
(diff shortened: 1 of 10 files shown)


KevinFairise2 (approved):
Just leave a small question, otherwise looks good to me!

recipes/repository.rb
```diff
@@ -174,14 +178,18 @@ def warn_deprecated_yumrepo_gpgkey
     action :remove
   end
 
-  apt_repository 'datadog_apt_A2923DFF56EDA6E76E55E492D3A80E30382E94DE' do
+  apt_repository 'datadog_apt_5F1E256061D813B125E156E8E6266D4AC0962C7D' do
```
bkabrda, line 181: No previous versions of the cookbook that were creating these `apt` repositories knew about this key, so we don't need to do this.

spec/repository_spec.rb
```diff
@@ -42,13 +44,55 @@
 
     expect(chef_run.remote_file('remote_file_DATADOG_APT_KEY_CURRENT.public')).to notify(
       'execute[import apt datadog key DATADOG_APT_KEY_CURRENT.public]').to(:run).immediately
+    expect(chef_run.remote_file('remote_file_5F1E256061D813B125E156E8E6266D4AC0962C7D')).to notify(
```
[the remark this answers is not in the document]
bkabrda: It will become the current key when we actually start using it (we're not using it yet).

spec/repository_spec.rb
```diff
@@ -42,13 +44,55 @@
...
+      'execute[import apt datadog key 5F1E256061D813B125E156E8E6266D4AC0962C7D]').to(:run).immediately
     expect(chef_run.remote_file('remote_file_D75CEA17048B9ACBF186794B32637D44F14F620E')).to notify(
       'execute[import apt datadog key D75CEA17048B9ACBF186794B32637D44F14F620E]').to(:run).immediately
     expect(chef_run.remote_file('remote_file_A2923DFF56EDA6E76E55E492D3A80E30382E94DE')).to notify(
       'execute[import apt datadog key A2923DFF56EDA6E76E55E492D3A80E30382E94DE]').to(:run).immediately
   end
 end
 
+def handle_keys(key_list, install_gnupg = true)
```
bkabrda, line 56: I don't feel like `handle_keys` is a very descriptive method name. Can you please improve it?

chouetz: I proposed `import_gpg_keys`, wdyt?

recipes/repository.rb
chouetz: So we add this during a next modification set when we remove old keys? I'm wondering if we could keep it if it's harmless?

bkabrda (approved):

PR #891, opened 2023-06-11
feat: support weblogic integration

vigohe:
### What does this PR do?

Support Weblogic integration

### DIFF
--- a/recipes/weblogic.rb
+++ b/recipes/weblogic.rb
@@ -1,0 +1,44 @@
+# Copyright:: 2011-Present, Datadog
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+include_recipe '::dd-agent'
+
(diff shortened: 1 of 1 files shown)


recipes/weblogic.rb
```diff
@@ -0,0 +1,42 @@
...
+#     :name => "test"
+#   }
+# ]
+
+datadog_monitor 'weblogic' do
+  instances node['datadog']['weblogic']['instances']
+  logs node['datadog']['weblogic']['logs']
+  action :add
+  notifies :restart, 'service[datadog-agent]' if node['datadog']['agent_start']
+end
```
KSerrania, line 42: Is there a specific reason not to use the default integration template, as is used in many other integrations, eg. aerospike: https://github.com/DataDog/chef-datadog/blob/4c39143a77ba608d6e1265c9a06a9f04497d5cc7/recipes/aerospike.rb#L431-L438?

Using the default integration template makes it so you don't have to define parsing rules in a specific erb file, and makes the recipe future-proof (if a new option is added to the `weblogic` integration in a future Agent version, the recipe won't need to be updated to be able to use it).

KSerrania (commented):
Thanks for opening this PR! I have one comment before approving.

recipes/weblogic.rb
vigohe: Sure I will change it

vigohe: Thanks, I did change the code

KSerrania (commented):
I left one last cleanup comment, otherwise LGTM.

templates/default/weblogic.yaml.erb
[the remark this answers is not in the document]
vigohe: Thanks, I remove the file!

chouetz (approved):
All comments integrated

PR #896, opened 2023-06-29
Update recipes to support amazonlinux 2023 installation

chouetz:
### What does this PR do?
Adapt recipes to work on Amazonlinux 2023 platform: yum cannot be used and standard stop/start/status commands where not found

### Motivation
Amazonlinux 2023 footprint is increasing in datacenters, we must ensure our installation methods support it

Install the agent on AmazonLinux2 (ami-04823729c75214919) and AmazonLinux2023 (ami-06b09bfacae1453cb)

### DIFF
--- a/libraries/recipe_helpers.rb
+++ b/libraries/recipe_helpers.rb
@@ -150,6 +150,7 @@ class Chef
             service_provider = Chef::Provider::Service.const_get(specified_provider)
           end
           service_provider
+        # Specific catch for Amazon Linux >= 2022 where Upstart doesn't work
         elsif systemd_platform?(node)
           Chef::Provider::Service::Systemd
         elsif upstart_platform?(node)


recipes/dd-agent.rb
```diff
@@ -140,6 +141,11 @@ def template_vars
 
 service 'datadog-agent' do
   service_name service_name
+  if is_amazon_linux_2023
+    status_command "systemctl status #{service_name}"
+    start_command "systemctl start #{service_name}"
+    stop_command "systemctl stop #{service_name}"
+  end
```
KSerrania, line 148: I'm surprised that this would be needed specifically for Amazon Linux 2023. Is Chef not able to resolve the correct service provider?

chouetz: On the test I ran yesterday I was facing this
```
    Error executing action `start` on resource 'service[datadog-agent]'

    Errno::ENOENT
    -------------
    No such file or directory - /sbin/start

```
So I fixed with this. And yes I was quite surprised, I can say what I did worked but I was puzzled to be forced to do that, so there is maybe something I did wrong

libraries/recipe_helpers.rb
```diff
@@ -146,6 +150,8 @@ def service_provider(node)
             service_provider = Chef::Provider::Service.const_get(specified_provider)
           end
           service_provider
+        elsif systemd_platform?(node)
```
chouetz, line 153: We catch Amazon Linux 2023 first, to use Systemd, and then default to the previous test without modification

recipes/dd-agent.rb
chouetz: Replaced with an update on recipe_helpers

libraries/recipe_helpers.rb
KSerrania: Can this be added as a comment in code?

KSerrania (approved):

chouetz:
Relates also to https://github.com/DataDog/chef-datadog/pull/523

PR #898, opened 2023-07-06
Support on old chef version

chouetz:
While doing QA for #874 I noticed 2 things:
- we cannot use more than one constraint on dependency specification (triggers  Chef::Exceptions::ObsoleteDependencySyntax)
- "yaml" format in [monitor resource](https://github.com/DataDog/chef-datadog/blob/main/resources/monitor.rb#L81) is not understood causing Chef::Exceptions::ChildConvergeError

### What does this PR do?
Adapt syntax to old chef versions (tested with 12.8.1)

### Motivation
Fix errors seen during releasing of 4.18.0 for PR #874 

Run the cookbook with chef-infra version 12.8.1

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -43,9 +43,10 @@ else
   depends 'apt'
 end
 
-# Use '< 5.0' with Chef < 12.14
+# Must be '>= 3.0' and '< 5.0' with Chef < 12.14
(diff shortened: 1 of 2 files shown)


amenasria (approved):
LGTM !

You just need to fix the style linter before merging.

About the `'~> 3.0' I hope it won't break anything though it seems fine.

PR #903, opened 2023-09-07
Add datadog_monitor option to override the config file name

brentm5:
### What does this PR do?

Allows users consuming the datadog_monitor resource to select a config file name instead of just the default of conf.yaml. The default still exists but by adding this option others can call the datadog_monitor function multiple times and not override existing files.

### Motivation

This allows people to use the integration multiple times and place different config files in the same integration directory.  The use case would be many different cookbooks adding a single check they care about without doing messy attribute overrides. The use case would be the following, Cookbook A & Cookbook B are both maintained by different teams but can be used on the same host.  In the previous way of doing this you would have to add attributes in different cookbooks which can cause weird issues around attribute inheritance / ordering.  By adding this teams can more easily just create different config files that are owned by the respective cookbooks and do not step on each other. 

Cookbook A: 

```ruby
# Cookbook A

datadog_monitor 'process' do
  action :add
   <Existing configuration for the monitor>
  config_name 'datadog-agent'
end

# Cookbook B

datadog_monitor 'process' do
  action :add
   <Existing configuration for the monitor>
  config_name 'ssh'
end
```

### Possible Drawbacks / Trade-offs

I don't see too many Drawbacks, by default the existing behavior exists.  The only drawback I can think of is this does add some complexity since there can be multiple files to manage however if you manage the with this resource then the delete function has the same logic. 

I am not entire sure what other tests should be added which is why I made this a Draft.  The testing should be light, essentially create an integration (logs might be the easiest since it doesn't require an integration to run) and just validate that the new file is generated.  The default case should already be tested. 

### Reviewer's Checklist

- [x] Add additional tests to validate the new behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,19 +320,21 @@ datadog_monitor 'name' do
   instances                         Array # default value: []
   logs                              Array # default value: []
   use_integration_template          true, false # default value: false
+  config_name                       String # default value: 'conf'
   action                            Symbol # defaults to :add
 end
 ```
 
 #### Properties
 
-| Property                   | Description                                                                                                                                                                                                                                                                                    |
-|----------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `'name'`                   | The name of the Agent integration to configure and enable.                                                                                                                                                                                                                                     |
-| `instances`                | The fields used to fill values under the `instances` section in the integration configuration file.                                                                                                                                                                                            |
-| `init_config`              | The fields used to fill values under the the `init_config` section in the integration configuration file.                                                                                                                                                                                      |
-| `logs`                     | The fields used to fill values under the the `logs` section in the integration configuration file.                                                                                                                                                                                             |
+| Property                   | Description                                                                                                                                                                                                                                                                                   |
+|----------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `'name'`                   | The name of the Agent integration to configure and enable.                                                                                                                                                                                                                                    |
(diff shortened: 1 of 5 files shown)


brentm5:
What are the steps to get a review or have a further discussion around the changes above?

spencergilbert:
Sorry @brentm5! This fell through the cracks, I've added this to our next sprint so we should be taking a look in the next 2-3 weeks.

alopezz (changes requested):
Thanks and apologies again for taking so long to look at this. This looks good implementation-wise, but it needs to be documented as otherwise customers may have trouble finding the feature.

I'm thinking something along the lines of adding references to it under the "Syntax" and "Properties" in https://github.com/DataDog/chef-datadog#integrations-without-recipes.

README.md
```diff
@@ -320,19 +320,21 @@ datadog_monitor 'name' do
...
+| `logs`                     | The fields used to fill values under the the `logs` section in the integration configuration file.                                                                                                                                                                                            |
 | `use_integration_template` | Set to `true` (recommended) to use the default template, which writes the values of `instances`, `init_config`, and `logs` in the YAML under their respective keys. This defaults to `false` for backward compatibility, but may default to `true` in a future major version of the cookbook. |
+| `config_name`              | The filename used when creating an integrations configuration file. Overriding this property allows the creation of multiple configuration files for a single integration.  This defaults to `conf` which creates a configuration file named `conf.yaml`.                                     |
```
jhgilbert, line 337:
```suggestion
| `config_name`              | The filename used when creating an integrations configuration file. Overriding this property allows the creation of multiple configuration files for a single integration.  This defaults to `conf`, which creates a configuration file named `conf.yaml`.                                     |
```

jhgilbert (approved):
Approved with a very minor suggestion, thanks!

alopezz (approved):
Cool, thanks!

PR #904, opened 2023-10-31
[Bug] File.exists return true when agent.exe doesn't exist but agent folder exist.

Pythyu:
### What does this PR do?

Fix a bug where `File.exists` return true when `agent.exe` doesn't exist but `agent` folder exist. Instead we're using `File.file` that checks if it's a file and if it exist.

### DIFF
--- a/libraries/recipe_helpers.rb
+++ b/libraries/recipe_helpers.rb
@@ -3,7 +3,6 @@
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-
 #     http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
--- a/libraries/recipe_helpers.rb
+++ b/libraries/recipe_helpers.rb
@@ -201,7 +201,7 @@ class Chef
 
         def fetch_current_version
           raw_version = agent_get_version
-          return nil if raw_version.nil?
+          return nil if raw_version.nil? or raw_version == ""
           match_data = raw_version.match(/^Agent ([^\s]*) (- Meta: ([^\s]*) )?- Commit/)
 
           version = match_data[1] if match_data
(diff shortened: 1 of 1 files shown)


chouquette (approved):
LGTM!

libraries/recipe_helpers.rb
```diff
@@ -3,7 +3,7 @@
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-#
+
```
amenasria:
```suggestion
#
```

libraries/recipe_helpers.rb
```diff
@@ -3,7 +3,6 @@
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-#
```
amenasria, line 6: **🥜 nitpick:** ‏Think we shouldn't modify this

libraries/recipe_helpers.rb
```diff
@@ -195,15 +194,15 @@ def must_reinstall?(node)
...
+          return nil unless File.file?(WIN_BIN_PATH)
           shell_out("\"#{WIN_BIN_PATH}\" version -n").stdout.strip
         end
 
         def fetch_current_version
           raw_version = agent_get_version
           return nil if raw_version.nil?
           match_data = raw_version.match(/^Agent ([^\s]*) (- Meta: ([^\s]*) )?- Commit/)
-
+          # will fail if raw_version is empty (Line 199 fails)
```
amenasria, line 205: **💬 suggestion:** ‏We shouldn't mention a line as the file keeps changing !

amenasria (approved):
Small nits otherwise LGTM!

libraries/recipe_helpers.rb
Pythyu: will edit the comment 👍

PR #907, opened 2024-01-24
add support for CWS on Windows

paulcacheux:
### What does this PR do?

This PR adds support for CWS on Windows. It also slightly cleans up the system probe recipe and adds a rspec test for this new support.

### Motivation

Windows agent builders use chef to install the agent, since we want to enable CWS there, we need this new support.

Use:
```
datadog:
  security_agent:
    cws:
      enabled: true
```
and check that CWS is actually enabled (this requires a custom agent build until 7.52 is released)

### DIFF
--- a/recipes/security-agent.rb
+++ b/recipes/security-agent.rb
@@ -20,11 +20,17 @@
 is_windows = platform_family?('windows')
 
 # Set the correct agent startup action
-security_agent_enabled = !is_windows && node['datadog']['security_agent']['cws']['enabled'] || node['datadog']['security_agent']['cspm']['enabled']
+security_agent_enabled = node['datadog']['security_agent']['cws']['enabled'] || (!is_windows && node['datadog']['security_agent']['cspm']['enabled'])
+security_agent_start = security_agent_enabled && node['datadog']['agent_start'] && node['datadog']['agent_enable'] ? :start : :stop
 
 #
 # Configures security-agent agent
-security_agent_config_file = '/etc/datadog-agent/security-agent.yaml'
+security_agent_config_file =
+  if is_windows
+    'C:/ProgramData/Datadog/security-agent.yaml'
+  else
+    '/etc/datadog-agent/security-agent.yaml'
+  end
 security_agent_config_file_exists = ::File.exist?(security_agent_config_file)
 
 template security_agent_config_file do
@@ -52,9 +58,11 @@ template security_agent_config_file do
     compliance_extra_config: compliance_extra_config
   )
 
-  owner 'root'
-  group 'dd-agent'
-  mode '640'
+  unless is_windows
+    owner 'root'
+    group 'dd-agent'
+    mode '640'
+  end
 
   notifies :restart, 'service[datadog-agent-security]', :delayed if security_agent_enabled
 
@@ -65,14 +73,18 @@ end
 # Common configuration
 service_provider = Chef::Datadog.service_provider(node)
 
-service_name = 'datadog-agent-security'
+service_name = is_windows ? 'datadog-security-agent' : 'datadog-agent-security'
 
-if security_agent_enabled
-  service 'datadog-agent-security' do
-    service_name service_name
-    action :start
-    provider service_provider unless service_provider.nil?
+service 'datadog-agent-security' do
+  service_name service_name
+  action [security_agent_start]
+  provider service_provider unless service_provider.nil?
+  if is_windows
+    supports :restart => true, :start => true, :stop => true
+    restart_command "powershell restart-service #{service_name} -Force"
+    stop_command "powershell stop-service #{service_name} -Force"
+  else
     supports :restart => true, :status => true, :start => true, :stop => true
-    subscribes :restart, "template[#{security_agent_config_file}]", :delayed
   end
+  subscribes :restart, "template[#{security_agent_config_file}]", :delayed if security_agent_enabled
 end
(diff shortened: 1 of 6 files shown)


derekwbrown (approved):

recipes/system-probe.rb
```diff
@@ -22,7 +22,7 @@
 # Set the correct agent startup action
 cws_enabled = node['datadog']['security_agent']['cws']['enabled']
 sysprobe_enabled = if is_windows
```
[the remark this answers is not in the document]
paulcacheux: I did something similar to what you suggested, I think it makes things easier anyway even if we still have a small `if is_windows`

recipes/system-probe.rb
```diff
@@ -20,12 +20,15 @@
...
+usm_enabled = node['datadog']['system_probe']['service_monitoring_enabled']
 cws_enabled = node['datadog']['security_agent']['cws']['enabled']
-sysprobe_enabled = if is_windows
-                     node['datadog']['system_probe']['network_enabled']
-                   else
-                     node['datadog']['system_probe']['enabled'] || node['datadog']['system_probe']['network_enabled'] || node['datadog']['system_probe']['service_monitoring_enabled'] || cws_enabled
-                   end
+
+sysprobe_enabled ||= npm_enabled || cws_enabled
+unless is_windows
```
derekwbrown, line 29: Windows has USM; so on windows system probe should be enabled if usm is enabled.

(this is also somewhat outside the scope of what you're doing, because it's clear the USM support for windows is lacking in the chef recipe).  So, I'm fine either way.  But there's no underlying reason for the conditional at all.

paulcacheux: Ah didn't know that, yeah the USM is checked here but not plugged anywhere after that. I will simplify further then

recipes/security-agent.rb
```diff
@@ -65,14 +73,18 @@
 # Common configuration
 service_provider = Chef::Datadog.service_provider(node)
 
-service_name = 'datadog-agent-security'
+service_name = is_windows ? 'datadog-security-agent' : 'datadog-agent-security'
```
chouetz, line 76: This is probably not related to this PR, but why???

spec/system-probe_spec.rb
```diff
@@ -240,4 +240,115 @@
...
+        }
+      end
+    end
+
+    cached(:chef_run) do
+      solo.converge(described_recipe) do
+        solo.resource_collection.insert(
+          Chef::Resource::Service.new('datadog-agent', solo.run_context))
+        solo.resource_collection.insert(
+          Chef::Resource::Service.new('datadog-agent-security', solo.run_context))
```
chouetz, line 329: If I believe line 76 in `recipes/security-agent.rb` in windows the name changes, correct?
```suggestion
          Chef::Resource::Service.new('datadog-security-agent', solo.run_context))
```

chouetz (approved):

recipes/security-agent.rb
```diff
@@ -20,11 +20,17 @@
 is_windows = platform_family?('windows')
 
 # Set the correct agent startup action
-security_agent_enabled = !is_windows && node['datadog']['security_agent']['cws']['enabled'] || node['datadog']['security_agent']['cspm']['enabled']
+security_agent_enabled = node['datadog']['security_agent']['cws']['enabled'] || (!is_windows && node['datadog']['security_agent']['cspm']['enabled'])
```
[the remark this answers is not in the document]
paulcacheux: I think the original code was wrong, CSPM does not support windows for now. So I'm moving the is windows check to only apply to the CSPM part

recipes/security-agent.rb
paulcacheux: I have no idea.. I'm just copying what has been done for system probe

spec/system-probe_spec.rb
paulcacheux: I think there is a difference between the "public name" of the service which changes on windows, and the name of the chef service which is common on both platforms

recipes/security-agent.rb
```diff
@@ -20,11 +20,17 @@
...
+security_agent_start = security_agent_enabled && node['datadog']['agent_start'] && node['datadog']['agent_enable'] ? :start : :stop
 
 #
 # Configures security-agent agent
-security_agent_config_file = '/etc/datadog-agent/security-agent.yaml'
+security_agent_config_file =
+  if is_windows
```
derekwbrown, line 29: this is probably OK for now, and is probably how we do it for system-probe.yaml, too.
But the root config dir is configurable on windows.
So, `c:\programdata...` is the correct default. But in theory it could be changed.  it's not clear if the chef recipe as a whole supports that.

recipes/security-agent.rb
derekwbrown: is the name of the service on linux `datadog-security-agent` or `datadog-agent-security`?
Right now, the name of the service (on windows) is `datadog-security-agent`.  Since we haven't shipped, we can make it `datadog-agent-security` without breaking anything, and then it will match linux if that's desirable.

derekwbrown (approved):
Approved as is; but noted there are some options we might want to take since we haven't shipped anything yet.

recipes/security-agent.rb
julien-lebot: We don't support changing the `PROJECTLOCATION` nor the `APPLICATIONDATADIRECTORY` with Chef on Windows.

derekwbrown: thanks.  that's even better.

PR #913, opened 2024-03-27
Fix failing CI, local dev dockerfile setup, implement Centos 6 deprecation

alopezz:
### What does this PR do?

- Implements deprecation of Centos <7 (cf. https://github.com/DataDog/ansible-datadog/pull/556, etc.).
- Makes the provided Dockerfile build successfully and capable of running kitchen-docker tests locally (at least on Intel).
- Updates documentation related to that setup
- Updates machine image used for kitchen-docker-tests job, since that fixes some failing tests ([example](https://app.circleci.com/pipelines/github/DataDog/chef-datadog/888/workflows/840e772f-7ea0-4fa9-8f42-0c20e474bf64/jobs/10637)).

### Motivation

[BARX-282](https://datadoghq.atlassian.net/browse/BARX-282)

### Additional Notes

- I tested the "erroring out on invalid pinned version" manually because I don't feel too confident on implementing those tests myself at this point.
- I'd have rather split this in several PR's into several since it's tackling a few concerns at the same time, but I had no choice if I wanted to test things locally and the CI green.
- Tests *seem* to work fine on Apple silicon (M1s, etc.), though they're expected to run more slowly due to the extra emulation layer.

### Reviewer's Checklist

[BARX-282]: https://datadoghq.atlassian.net/browse/BARX-282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #914, opened 2024-04-10
Remove centos 6 kitchen tests

Pythyu:
### What does this PR do?

Remove centos 6 tests following centos 6 deprecation and fixing main. Will need to be re-introduced into centos 6 pinned version test later (prbly via #913)

### DIFF
--- a/kitchen.docker.yml
+++ b/kitchen.docker.yml
@@ -18,14 +18,6 @@ platforms:
   - name: ubuntu-14.04
     driver_config:
       require_chef_omnibus: 14.12
-  - name: centos-6.6
-    driver_config:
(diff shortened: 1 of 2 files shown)


chouquette (approved):
LGTM! Did you create a card for the "Will need to be reintroduced later" part so we don't forgot?

alopezz:
FYI I'm working on this: https://github.com/DataDog/chef-datadog/pull/913.

I'm also surprised that removing centos-6 tests fixes the centos-7 tests (which I'm trying to fix on that PR).

Pythyu:
Yea I saw @alopezz'PR that's why I expect him to re-introduce the test once fixed. Just to unlock main for now 👍

PR #916, opened 2024-05-17
[2024 GPG Key Rotation] Import new GPG future key 

Pythyu:
### What does this PR do?

Import a new future GPG key 
Also fix main pipelines that are broken due to non-pinned dependency

### Motivation

WIth the 2024 GPG key rotation ongoing customer will need to start trusting a new future GPG key that will be used in 2028.

PR #918, opened 2024-06-20
Datadog gpg

vakpande:


rivlinpereira (approved):

Pythyu (commented):
Hey 👋 
We are already importing new datadog gpg keys since this [PR](https://github.com/DataDog/chef-datadog/pull/916/files). Unless I'm missing some context (if so please add it to the PR description 🙏 ) we shouldn't need to add anything to import apt keys

PR #924, opened 2024-07-10
Fix dry-inflector dependency

AliDatadog:


Pythyu (approved):
LGTM we'll need to fix the kitchen tests in a separate PR

